PR #398, opened 2022-11-03
set default low-rank initializer explicitly

MUCDK:
At the moment, the initializer has to be set explicitly when using low rank in the `TemporalProblem` and the `LineageProblem`. As we observed that `rank2` performs best in general, we here explicitly set the LR initializer to `rank2`


src/moscot/_docs/_docs.py
```diff
@@ -202,7 +202,7 @@
         - `k-means` :cite:`scetbon:22b`
         - `generalized-k-means` :cite:`scetbon:22b`
 
-    If `None`, the default for not low rank is `default`, for low rank it is `random`.
+    If `None`, the default for not low rank is `default`, for low rank it is `rank2`.
```
giovp, line 205:
```suggestion
    If `None`, the default is `default`. If `rank != -1`, the default is `rank2`.
```

tests/problems/conftest.py
```diff
@@ -114,7 +114,7 @@ def adata_time_with_tmap(adata_time: AnnData) -> AnnData:
     "scale_cost": "max_cost",
     "rank": -1,
     "batch_size": 122,
-    "initializer": "quad_initializer",
+    "initializer": None,  # full rank only suppo
```
giovp, line 117:
```suggestion
    "initializer": None,
```

giovp (approved):

Issue #367, opened 2022-10-29
pass custom cost


giovp:
this could be an intuitive API

```python
tp = TemporalProblem(adata)
tp = tp.prepare([1, 2, 3, 4])
tp[0, 2].set_cost(cost_matrix, "custom_matrix")
```
and this would modify the TaggedArray in place like this.
```
TaggedArray(tag="pc", data, data_y) -> TaggedArraY(tag="cost", data=cost_matrix, data_y =None)
```

@michalk8 @MUCDK ideas? This would be done only after release (but before paper).

MUCDK:
closed by #414

PR #414, opened 2022-11-22
Feature/set custom cost

MUCDK:


MUCDK:
closes #367

src/moscot/problems/base/_base_problem.py
```diff
@@ -474,6 +475,120 @@ def _create_marginals(
     def _estimate_marginals(self, adata: AnnData, *, source: bool, **kwargs: Any) -> ArrayLike:
         return np.ones((adata.n_obs,), dtype=float) / adata.n_obs
 
+    @d.dedent
+    def set_xy(
```
giovp, line 479: I think considering the rather the potential unintuitive behaviour, I would __always__ validate the data. What's a use case where this is not true?

tests/problems/base/test_general_problem.py
```diff
@@ -43,6 +45,92 @@ def test_passing_scale(self, adata_x: AnnData, scale_cost: str):
 
         np.testing.assert_allclose(gt.matrix, sol.transport_matrix, rtol=RTOL, atol=ATOL)
 
+    @pytest.mark.parametrize("validate_data", [True, False])
+    def test_set_xy(self, adata_x: AnnData, adata_y: AnnData, validate_data: bool):
+        rng = np.random.RandomState(42)
```
giovp, line 50: I'd create a RNG up at the top of this file and reuse it.

tests/problems/base/test_general_problem.py
```diff
@@ -43,6 +45,92 @@ def test_passing_scale(self, adata_x: AnnData, scale_cost: str):
...
+        )
+
+        cm = rng.uniform(1, 10, size=(adata_x.n_obs, adata_y.n_obs))
+        cost_matrix = pd.DataFrame(index=adata_x.obs_names, columns=adata_y.obs_names, data=cm)
+        prob.set_xy(cost_matrix, tag="cost", validate_data=validate_data)
+
+        prob = prob.solve(max_iterations=5)
+        np.testing.assert_equal(prob.xy.data_src, cost_matrix.to_numpy())
+
+    @pytest.mark.parametrize("validate_data", [True, False])
```
giovp, line 65: kernel is never tested.

src/moscot/problems/base/_base_problem.py
```diff
@@ -474,6 +475,120 @@ def _create_marginals(
...
+            logger.info(f"Changing the problem type from {self.problem_kind} to fused-quadratic.")
+            self._problem_kind = ProblemKind.QUAD_FUSED
+        if data.shape != (self.adata_src.n_obs, self.adata_tgt.n_obs):
+            raise ValueError(
+                f"`data` is exptected to have shape {(self.adata_src.n_obs, self.adata_tgt.n_obs)} but found {data.shape}."  # noqa: E501
+            )
+        if validate_data:
+            if not isinstance(data, pd.DataFrame):
+                raise TypeError("If the data is to be validated, the data must be of type pandas.DataFrame.")
+            if set(data.index) != set(self.adata_src.obs_names):
```
giovp, line 505: why set? shouldn't it be exactly the same?

src/moscot/problems/base/_base_problem.py
```diff
@@ -474,6 +475,120 @@ def _create_marginals(
...
+        %(data_set)s
+        %(tag)s
+        %(validate_data_set)s
+
+        Returns
+        -------
+        None
+        """
+        if self.problem_kind == ProblemKind.QUAD:
+            logger.info(f"Changing the problem type from {self.problem_kind} to fused-quadratic.")
```
giovp, line 496: I don't think we should allow to change the method from linear to fused quadratic after initialization. Or maybe I'm not seeing it, could you share a use case?

giovp (changes requested):

src/moscot/problems/base/_base_problem.py
MUCDK: A setting would be to have the `TemporalProblem` and then you want to have a prior. Example: Use gene expression in the linear term, and now you want to have ATAC in the quadratic term. The `TemporalProblem` is not constructed to explicitly allow for this setting (as ATAC lives in the same space it is actually not necessary), but the use might want to play around with it.

src/moscot/problems/base/_base_problem.py
MUCDK: If you do not have exactly the same cell names but you are certain that the order is correct. As by default validation is required, I think it is fair to keep it this way. Do you have a strong opinion?

PR #418, opened 2022-11-24
pass cost

MUCDK:


MUCDK:
After merging #414 I would use `set_x/y` as a standard way to pass custom cost matrices (as opposed to `obsp`). Hence, I would also change the default for the generic `GWProblem` and `FGWProblem` to pointclouds.

giovp (approved):

PR #423, opened 2022-11-30
Feature/correlation test

MUCDK:


src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +477,93 @@ def _cell_aggregation_transition(
...
+    ) -> pd.DataFrame:
+        """
+        Compute correlation of push or pull distribution with features.
+
+        Correlates (processed) count data with probabilities of cells mapped to a subset of cells, e.g. for a
+        certain celltype.
+
+        Parameters
+        ----------
+        key
```
giovp, line 502: maybe call it `obs_key`

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +477,93 @@ def _cell_aggregation_transition(
...
+        %(correlation_test.returns)s
+        """
+        if key not in self.adata.obs.columns:
+            raise KeyError(f"[{key!r} not found in `adata.obs`.")
+
+        method = TestMethod(method)
+
+        if annotation is not None:
+            annotation_key, annotation_vals = next(iter(annotation.items()))  # type:ignore[misc]
+            if annotation_key not in self.adata.obs.columns:  # type:ignore[has-type]
```
giovp, line 538:
```suggestion
            if annotation_key not in self.adata.obs:  # type:ignore[has-type]
```

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +477,93 @@ def _cell_aggregation_transition(
...
+        else:
+            adata_red = self.adata
+
+        adata_red = adata_red[~adata_red.obs[key].isnull()]
+        if len(adata_red) == 0:
+            raise ValueError(f"`adata.obs[{key!r}]` only contains NaN values.")
+        distribution = adata_red.obs[[key]]
+
+        if features is not None:
+            adata_red = adata_red[:, features]
```
giovp, line 553: I would use `sc.get.obs_df` so that you can be flexible with either features or obs

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +477,93 @@ def _cell_aggregation_transition(
...
+
+        if layer is not None and layer not in self.adata.layers:
+            raise KeyError(f"Layer `{layer!r}` not found in `adata.layers`.")
+
```
giovp, line 557: `sc.get.obs_df` would also handle this in a nicer way

src/moscot/problems/base/_utils.py
```diff
@@ -228,3 +236,242 @@ def _filter_kwargs(*funcs: Callable[..., Any], **kwargs: Any) -> Dict[str, Any]:
...
+        res[f"{c}_ci_low"] = ci_low[:, idx]
+        res[f"{c}_ci_high"] = ci_high[:, idx]
+
+    # fmt: off
+    res = res[[f"{c}_{stat}" for c in Y.columns for stat in ("corr", "pval", "qval", "ci_low", "ci_high")]]
+    return res.sort_values(by=[f"{c}_corr" for c in Y.columns], ascending=False)
+    # fmt: on
+
+
+def _correlation_test_helper(
```
giovp, line 327: this is all taken frm cellrank?

giovp (commented):
first pass. Is the whole testing + computation of correlation from cellrank? Can't we use like scipy for that?

src/moscot/problems/_subset_policy.py
```diff
@@ -179,7 +179,7 @@ def remove_node(self, node: Tuple[K, K]) -> None:
             pass
 
 
-class OrderedPolicy(SubsetPolicy[K], ABC):
```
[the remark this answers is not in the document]
MUCDK: yes, and actually don't see a reason why it should be ABC

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +477,93 @@ def _cell_aggregation_transition(
             tm = pd.concat([tm, to_app], verify_integrity=True, axis=0)
             df_1 = df_1.drop(current_cells, axis=1)
         return tm
+
+    # adapted from CellRank (github.com/theislab/cellrank)
+    def compute_feature_correlation(
+        self: AnalysisMixinProtocol[K, B],
+        key: str = None,
```
MUCDK, line 484: remove default.

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +477,93 @@ def _cell_aggregation_transition(
...
+        annotation: Optional[Dict[str, List[str]]] = None,
+        layer: Optional[str] = None,
+        features: Optional[List[str]] = None,
+        confidence_level: float = 0.95,
+        n_perms: int = 1000,
+        seed: Optional[int] = None,
+        **kwargs: Any,
+    ) -> pd.DataFrame:
+        """
+        Compute correlation of push or pull distribution with features.
```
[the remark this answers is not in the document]
MUCDK: Currently, only implemented for one key / one distribution

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +477,93 @@ def _cell_aggregation_transition(
...
+
+        Correlates (processed) count data with probabilities of cells mapped to a subset of cells, e.g. for a
```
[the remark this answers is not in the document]
MUCDK: well would be some processed count data , right?

src/moscot/problems/base/_utils.py
MUCDK: yes, slightly adapted.

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +477,93 @@ def _cell_aggregation_transition(
...
+
+        Parameters
+        ----------
+        key
+            Column of :attr:`adata.obs` containing push-forward or pull-back distributions.
+        method
+            Mode to use when calculating p-values and confidence intervals. Valid options are:
+                - `{TestMethod.FISCHER!r}` - use Fischer transformation :cite:`fischer:21`.
+                - `{TestMethod.PERM_TEST!r}` - use permutation test.
+        annotation
```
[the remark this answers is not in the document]
MUCDK: Probably you refer to `source_groups` in `def cell_transition`. The data type is the same, the docs too specific to use it here.

src/moscot/problems/base/_mixins.py
MUCDK: but this would only give the `pd.DataFrame`, right

giovp: yes but I think that'd be ok cause then you have the flexibility of taking also continuous variables from obs

src/moscot/problems/base/_mixins.py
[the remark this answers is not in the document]
giovp: what if in `adata.obs` I have continuous time progression or pseudotime or deconvolution results? think that there are many cases where this applied.

src/moscot/problems/base/_mixins.py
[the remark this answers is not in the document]
giovp: yes, ah ok

src/moscot/problems/_subset_policy.py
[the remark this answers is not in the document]
michalk8: `_create_graph` is missing in both.

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+        Correlates (processed) count data with probabilities of cells mapped to a subset of cells, e.g. for a
+        certain celltype.
+
+        Parameters
+        ----------
+        obs_key
+            Column of :attr:`adata.obs` containing push-forward or pull-back distributions.
+        method
+            Mode to use when calculating p-values and confidence intervals. Valid options are:
+                - `{CorrTestMethod.FISCHER!r}` - use Fischer transformation :cite:`fischer:21`.
```
michalk8, line 507: Don't think the list renders nicely in the docs (also in other cases).

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+        features
+            Features of :class:`anndata.AnnData` which the correlation of ``anndata.AnnData.obs['{obs_key}']`` is
+            computed with:
+            - If `None`, all features will be taken into account.
+            - If of type :obj:`list`, features from :attr:`anndata.AnnData.var_names` will be taken.
+            - If `human`, `mouse`, or `drosophila`, the features are subsetted to transcription factors.
+        confidence_level
+            Confidence level for the confidence interval calculation. Must be in interval `[0, 1]`.
+        n_perms
+            Number of permutations to use when ``method = {CorrTestMethod.PERM_TEST!r}``.
```
michalk8, line 526: Change the ``method = {CorrTestMethod.PERM_TEST!r}`` please, it's no longer extrapolated (it was in my code since I was using the decorator for this).

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
             tm = pd.concat([tm, to_app], verify_integrity=True, axis=0)
             df_1 = df_1.drop(current_cells, axis=1)
         return tm
+
+    # adapted from CellRank (github.com/theislab/cellrank)
+    def compute_feature_correlation(
```
michalk8, line 483: Missing `@d.dedent`

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+        seed
+            Random seed when ``method = {CorrTestMethod.PERM_TEST!r}``.
+
+        %(parallel_kwargs)s
+
+        Returns
+        -------
+        %(correlation_test.returns)s
+        """
+        if obs_key not in self.adata.obs.columns:
```
michalk8, line 536: No need to access `.columns`.

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+            raise KeyError(f"[{obs_key!r} not found in `adata.obs`.")
```
michalk8, line 537: KeyError("Unable to access data in \`adata.obs[{obs_key!r}]\`.").

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+
+        method = CorrTestMethod(method)
+
+        if annotation is not None:
+            annotation_key, annotation_vals = next(iter(annotation.items()))
+            if annotation_key not in self.adata.obs:
+                raise KeyError(f"[{annotation_key!r} not found in `adata.obs`.")
+            if not isinstance(annotation_vals, list):
```
michalk8, line 545: The `list` restriction seems too much.

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+                raise TypeError("`annotation` expected to be dictionary of length 1 with value being a list.")
+
+            adata_red = self.adata[self.adata.obs[annotation_key].isin(annotation_vals)]
+        else:
+            adata_red = self.adata
+
+        adata_red = adata_red[~adata_red.obs[obs_key].isnull()]
+        if len(adata_red) == 0:
```
michalk8, line 553: Use `adata_red.n_obs`.

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+            raise ValueError(f"`adata.obs[{obs_key!r}]` only contains NaN values.")
+        distribution = adata_red.obs[[obs_key]]
+
+        if isinstance(features, str):
+            tfs = TranscriptionFactors.transcription_factors(organism=features)
+            adata_red = adata_red[:, list(set(tfs).intersection(adata_red.var_names))]
```
michalk8, line 559: Afterwards, you never check whether `adata.n_var` is `0` - does it produce a meaningful message if it errors? If not, please check and add one.

src/moscot/problems/base/_mixins.py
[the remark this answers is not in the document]
michalk8: I agree with Giovanni, should be more general (and the counts can be used as an example/illustration).

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+        """
+        Compute correlation of push or pull distribution with features.
+
+        Correlates (processed) count data with probabilities of cells mapped to a subset of cells, e.g. for a
+        certain celltype.
+
+        Parameters
+        ----------
+        obs_key
+            Column of :attr:`adata.obs` containing push-forward or pull-back distributions.
```
michalk8, line 504: :attr:\`anndata.AnnData.obs\`

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+            - If of type :obj:`list`, features from :attr:`anndata.AnnData.var_names` will be taken.
+            - If `human`, `mouse`, or `drosophila`, the features are subsetted to transcription factors.
+        confidence_level
+            Confidence level for the confidence interval calculation. Must be in interval `[0, 1]`.
+        n_perms
+            Number of permutations to use when ``method = {CorrTestMethod.PERM_TEST!r}``.
+        seed
+            Random seed when ``method = {CorrTestMethod.PERM_TEST!r}``.
+
+        %(parallel_kwargs)s
```
michalk8, line 530: This is not being interpolated atm (along with other other params).

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+        method = CorrTestMethod(method)
+
+        if annotation is not None:
+            annotation_key, annotation_vals = next(iter(annotation.items()))
+            if annotation_key not in self.adata.obs:
+                raise KeyError(f"[{annotation_key!r} not found in `adata.obs`.")
+            if not isinstance(annotation_vals, list):
+                raise TypeError("`annotation` expected to be dictionary of length 1 with value being a list.")
+
+            adata_red = self.adata[self.adata.obs[annotation_key].isin(annotation_vals)]
```
michalk8, line 548: Pet peeve: why not just call this `adata`? The `_red` suffixes evokes a color 🙈 .

src/moscot/problems/base/_mixins.py
```diff
@@ -476,3 +478,100 @@ def _cell_aggregation_transition(
...
+        else:
+            adata_red = self.adata
+
+        adata_red = adata_red[~adata_red.obs[obs_key].isnull()]
+        if len(adata_red) == 0:
+            raise ValueError(f"`adata.obs[{obs_key!r}]` only contains NaN values.")
+        distribution = adata_red.obs[[obs_key]]
```
michalk8, line 555: Add a comment please that you expect `distribution` to be a data frame instead of a series.

src/moscot/problems/base/_utils.py
```diff
@@ -228,3 +236,242 @@ def _filter_kwargs(*funcs: Callable[..., Any], **kwargs: Any) -> Dict[str, Any]:
...
+    Perform a statistical test for correlation between X and .
+
+    Return NaN for genes which don't vary across cells.
+
+    Parameters
+    ----------
+    X
+        Array or sparse matrix of shape ``(n_cells, n_features)`` containing the counts.
+    Y
+        Data frame of shape ``(n_cells, 1)`` containing the pull/push distribution.
```
michalk8, line 262: Afaik in the `CellRank` code, this can be `(n_cells, n_lineages)`, why only 1? It will have worse performance.

src/moscot/problems/base/_mixins.py
michalk8: Come to think of this, why it's `(n_cells, 1)` only? There can be many things correlated in the original implementation (this slightly limits the use of `obs_key`, which must be a string, but it could be a sequence of strings in `adata.obs`).

src/moscot/problems/base/_utils.py
```diff
@@ -228,3 +236,242 @@ def _filter_kwargs(*funcs: Callable[..., Any], **kwargs: Any) -> Dict[str, Any]:
...
+    method
+        Method for p-value calculation.
+    n_perms
+        Number of permutations if ``method='perm_test'``.
+    seed
+        Random seed if ``method = 'perm_test'``.
+    confidence_level
+        Confidence level for the confidence interval calculation. Must be in `[0, 1]`.
+
+    %(parallel_kwargs)s
```
michalk8, line 354: Again not interpolated, though not as much of a problem since this is a private function.

src/moscot/problems/base/_utils.py
```diff
@@ -228,3 +236,242 @@ def _filter_kwargs(*funcs: Callable[..., Any], **kwargs: Any) -> Dict[str, Any]:
...
+    ixs: ArrayLike,
+    corr: ArrayLike,
+    X: Union[ArrayLike, spmatrix],
+    Y: ArrayLike,
+    seed: Optional[int] = None,
+    queue=None,
+) -> Tuple[ArrayLike, ArrayLike]:
+    rs = np.random.RandomState(None if seed is None else seed + ixs[0])
+    cell_ixs = np.arange(X.shape[1])
+    pvals = np.zeros_like(corr, dtype=np.float64)
```
michalk8, line 458: Might need some cleaning-up, as far as I remember, `np.float64` alias has been deprecated.

src/moscot/problems/generic/_mixins.py
```diff
@@ -86,13 +89,124 @@ def cell_transition(
...
+            assert isinstance(result, dict)
+
+        if key_added is not None:
+            if TYPE_CHECKING:
+                assert isinstance(key_added, str)
+            plot_vars = {
+                "distribution_key": self.batch_key,
+            }
+            self.adata.obs[key_added] = self._flatten(result, key=self.batch_key)
+            Key.uns.set_plotting_vars(self.adata, AdataKeys.UNS, PlottingKeys.PUSH, key_added, plot_vars)
```
michalk8, line 145: Why does `set_plotting_vars` need the `AdataKeys.UNS`?

michalk8 (changes requested):

src/moscot/problems/base/_mixins.py
michalk8: In `CellRank`, I only return 1 dataframe that has many columns (`k x 5`, where `k` is the number of columns in `y` and `5` is because of the corr, p-val, q-val, ci-low, ci-high).

src/moscot/problems/base/_mixins.py
MUCDK: wanna check for an iterable, e.g. `if not isinstance(annotation_vals, collections.abc.Iterable)`?

src/moscot/problems/base/_mixins.py
MUCDK: doing it for you!

src/moscot/problems/base/_utils.py
MUCDK: just checked and didn't find any indication it was deprecated, see e.g. https://numpy.org/doc/stable/user/basics.types.html

src/moscot/problems/generic/_mixins.py
MUCDK: this is `moscot_results`, we save everything there. Or what do you mean?

src/moscot/problems/base/_mixins.py
MUCDK: Then it becomes more complicated with sorting, because then by default it's sorted w.r.t. a certain correlation value which might be misleading IMO. Don't you think so?

Open to have `obs_key` as a list but might be slightly too complicated I think. If you have a strong opinion on allowing it, I can implement it.

src/moscot/problems/base/_utils.py
michalk8: You're right, my mistake, these are the [deprecated aliases](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated) (e.g., `np.float`).

src/moscot/problems/generic/_mixins.py
michalk8: Since you call `Key.uns`, it seems like double info is used (i.e., it should always use `AdataKeys.UNS`), right? Would then remove this argument, but maybe better for another PR.

michalk8 (changes requested):
Only small changes, then LGTM.

PR #426, opened 2022-12-08
adapt callbacks and rename tag `cost` to `cost_matrix`

MUCDK:


src/moscot/problems/base/_compound_problem.py
```diff
@@ -608,16 +642,18 @@ def _cost_matrix_callback(
...
             if issparse(linear_cost_matrix):
                 linear_cost_matrix = linear_cost_matrix.A
             return {"xy": TaggedArray(linear_cost_matrix, tag=Tag.COST_MATRIX)}
 
-        return {
-            "x": TaggedArray(data[src_mask, :][:, src_mask], tag=Tag.COST_MATRIX),
-            "y": TaggedArray(data[tgt_mask, :][:, tgt_mask], tag=Tag.COST_MATRIX),
-        }
+        if term in ("x", "y"):
+            self._policy.create_mask(key_1, allow_empty=False)
```
michalk8, line 656: No need to call `create_mask` again.

michalk8 (changes requested):

src/moscot/problems/base/_base_problem.py
```diff
@@ -452,14 +458,13 @@ def concat(x: ArrayLike, y: ArrayLike) -> ArrayLike:
...
-        logger.info(f"Computing pca with `n_comps={n_comps}` for `x` and `y` using `{msg}`")
-        x = sc.pp.pca(x, n_comps=n_comps, **kwargs)
-        y = sc.pp.pca(y, n_comps=n_comps, **kwargs)
-        if scaler is not None:
-            x = scaler.fit_transform(x)
-            y = scaler.fit_transform(y)
-        return {"x": TaggedArray(x, tag=Tag.POINT_CLOUD), "y": TaggedArray(y, tag=Tag.POINT_CLOUD)}
+        if term in ("x", "y"):
+            logger.info(f"Computing pca with `n_comps={n_comps}` for `{term}` using `{msg}`")
+            x = sc.pp.pca(x, n_comps=n_comps, **kwargs)
```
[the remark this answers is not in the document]
MUCDK: Yes, if we don't have a shared space, then `adata_y` is always `None`. I thought it's easier to do so because then `adata` in the function definition can be made non-optional. 

Now we just call this function for each term`x`, `y`, and `xy` once.

src/moscot/problems/_utils.py
```diff
@@ -85,6 +83,26 @@ def handle_joint_attr(
         }
         return xy, kwargs
     if isinstance(joint_attr, Mapping):
+        joint_attr = dict(joint_attr)
+        if "attr" in joint_attr and joint_attr["attr"] == "X":
+            return {"x_attr": "X", "y_attr": "X"}, kwargs
+        if "attr" in joint_attr and joint_attr["attr"] == "obsm":
+            if "key" not in joint_attr:
```
michalk8, line 90: Just a small note, no need to change anything: slightly more "Pythonic" way would be to
ask for forgiveness rather than a permission as in:
```python
try:
    xy = ...
except KeyError:
    raise KeyError("`key` must be provided when ...")

File: src/moscot/problems/_utils.py [hunk 4]
```diff
@@ -85,6 +83,26 @@ def handle_joint_attr(
         }
         return xy, kwargs
     if isinstance(joint_attr, Mapping):
+        joint_attr = dict(joint_attr)
```
michalk8, line 86: Would be nice to have a small comment explaining the logic (only briefly).

src/moscot/problems/base/_base_problem.py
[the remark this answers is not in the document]
michalk8: Ok, thanks for the explanation, please add a comment similar to this in the code.

src/moscot/problems/base/_compound_problem.py
```diff
@@ -143,11 +150,25 @@ def _create_problems(
                 tgt_name = tgt
 
             problem = self._create_problem(src, tgt, src_mask=src_mask, tgt_mask=tgt_mask)
-            if callback is not None:
-                data = self._callback_handler(src, tgt, problem, callback=callback, **callback_kwargs)
-                kws = {**kwargs, **data}  # type: ignore[arg-type]
+            if xy_callback is not None:
+                xy_data = self._callback_handler(
```
michalk8, line 154: Can you please refactor this a bit please? Maybe allow `callback` in `_callback_handler` to be `None` and then return `{}` in this case.

src/moscot/problems/base/_compound_problem.py
```diff
@@ -608,16 +648,19 @@ def _cost_matrix_callback(
...
-        if return_linear:
-            linear_cost_matrix = data[src_mask, :][:, tgt_mask]
+        if term == "xy":
+            if key_2 is None:
+                raise ValueError("If `term` is `xy`, `key_2` cannot be `None`.")
+            mask_2 = self._policy.create_mask(key_2, allow_empty=False)
+
+            linear_cost_matrix = data[mask, :][:, mask_2]
             if issparse(linear_cost_matrix):
                 linear_cost_matrix = linear_cost_matrix.A
```
michalk8, line 660: Consider adding a warning here that a matrix is being densified.

michalk8 (changes requested):

PR #431, opened 2022-12-14
Simplify linear operator

michalk8:
No longer takes `forward` as an argument

### DIFF
--- a/src/moscot/problems/space/_mixins.py
+++ b/src/moscot/problems/space/_mixins.py
@@ -52,13 +52,13 @@ class SpatialAlignmentMixinProtocol(AnalysisMixinProtocol[K, B]):
 
     @staticmethod
     def _affine(  # type:ignore[empty-body]
-        tmap: LinearOperator, tgt: ArrayLike, src: ArrayLike, *, forward: bool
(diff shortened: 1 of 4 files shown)


src/moscot/problems/space/_mixins.py
```diff
@@ -282,30 +276,28 @@ def _subset_spatial(
...
         return tgt, R
 
     @staticmethod
-    def _warp(tmap: LinearOperator, _: ArrayLike, src: ArrayLike, forward: bool = True) -> Tuple[ArrayLike, None]:
+    def _warp(tmap: LinearOperator, src: ArrayLike, tgt: ArrayLike) -> Tuple[ArrayLike, None]:
         """Warp transformation."""
-        if forward:
-            return tmap.dot(src), None
-        return tmap.T.dot(src), None
+        del tgt
```
giovp, line 299: do we need tgt here and in the signature?

giovp (changes requested):

src/moscot/problems/space/_mixins.py
michalk8: Yes, if I wanna still pass it as `kwargs`, if not, then no.

giovp (approved):

PR #432, opened 2022-12-14
fix return type in mpl

MUCDK:


MUCDK:
> Why the change of alias from `mpl -> mp`? Would revert to `mpl` since everyone uses it.

I/We normally abbreviate `moscot.plotting` with `mpl`. This causes issues when import both, e.g. in the tests. Do you prefer to change the abbr. of `moscot.plotting` instead?

michalk8:
> I/We normally abbreviate `moscot.plotting` with `mpl`. This causes issues when import both, e.g. in the tests. Do you prefer to change the abbr. of `moscot.plotting` instead?

Would suggest using `import moscot as msc; msc.pl`, but am curious to hear from others (@giovp @zoepiran).

zoepiran:
> > I/We normally abbreviate `moscot.plotting` with `mpl`. This causes issues when import both, e.g. in the tests. Do you prefer to change the abbr. of `moscot.plotting` instead?
> 
> Would suggest using `import moscot as msc; msc.pl`, but am curious to hear from others (@giovp @zoepiran).


I like `import moscot as msc; msc.pl` ! :)

PR #437, opened 2022-12-23
Remove `FGWSolver`

### LINKED ISSUE #436: Remove `FGWSolver` in favor of `GWSolver`

michalk8:
closes #436

### DIFF
--- a/src/moscot/backends/ott/_solver.py
+++ b/src/moscot/backends/ott/_solver.py
@@ -275,10 +275,10 @@ class GWSolver(OTTJaxSolver):
             raise ValueError(f"Unable to create geometry from `x={x}`, `y={y}`.")
         geom_xx = self._create_geometry(x, **kwargs)
         geom_yy = self._create_geometry(y, **kwargs)
-        if alpha == 0 or xy is None:
-            # arbitrary positive fused penalty
+        if alpha == 1.0 or xy is None:  # GW
+            # arbitrary fused penalty (must be positive)
             geom_xy, fused_penalty = None, 1.0
-        else:
+        else:  # FGW
             fused_penalty = self._alpha_to_fused_penalty(alpha)
             geom_xy = self._create_geometry(xy, **kwargs)
             self._validate_geoms(geom_xx, geom_yy, geom_xy)
--- a/src/moscot/backends/ott/_solver.py
+++ b/src/moscot/backends/ott/_solver.py
@@ -23,7 +23,7 @@ from moscot.problems.base._utils import _filter_kwargs
 from moscot.solvers._base_solver import OTSolver, ProblemKind
 from moscot.solvers._tagged_array import TaggedArray
 
-__all__ = ["OTTCost", "SinkhornSolver", "GWSolver", "FGWSolver"]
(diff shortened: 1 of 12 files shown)


src/moscot/backends/ott/_solver.py
```diff
@@ -268,18 +268,40 @@ def _prepare(
...
         **kwargs: Any,
     ) -> QuadraticProblem:
-        del xy
         if x is None or y is None:
             raise ValueError(f"Unable to create geometry from `x={x}`, `y={y}`.")
-        geom_x = self._create_geometry(x, **kwargs)
-        geom_y = self._create_geometry(y, **kwargs)
+        geom_xx = self._create_geometry(x, **kwargs)
+        geom_yy = self._create_geometry(y, **kwargs)
+        if alpha == 0 or xy is None:
```
MUCDK, line 278: shouldn't we check for `alpha==1`? `alpha==0` should not be allowed as this corresponds to pure Sinkhorn.

src/moscot/backends/ott/_solver.py
```diff
@@ -268,18 +268,40 @@ def _prepare(
...
+            # arbitrary positive fused penalty
+            geom_xy, fused_penalty = None, 1.0
```
MUCDK, line 280: can we set `fused_penalty` to 0.0 then? this makes more sense I think.

src/moscot/problems/generic/_generic.py
```diff
@@ -260,6 +262,7 @@ def prepare(
     @d.dedent
     def solve(
         self,
+        alpha: float = 0.5,
```
MUCDK, line 265: can we set it to `1` by default, please? I.e. having pure `GW` per default.

tests/problems/generic/test_fgw_problem.py
```diff
@@ -27,7 +27,7 @@ class TestFGWProblem:
     @pytest.mark.fast()
     def test_prepare(self, adata_space_rotate: AnnData):
         expected_keys = [("0", "1"), ("1", "2")]
-        problem = FGWProblem(adata=adata_space_rotate)
```
MUCDK, line 30: When changing default `alpha` to 1 we will have to set `alpha` in this file explicitly in all tests.

MUCDK (changes requested):
Thanks!

src/moscot/backends/ott/_solver.py
michalk8: You're right, thanks for catching this!

src/moscot/backends/ott/_solver.py
michalk8: It has to be positive, since we check for this in `ott`; updated the comment.

src/moscot/problems/generic/_generic.py
michalk8: Done.

tests/problems/generic/test_fgw_problem.py
michalk8: Done.

MUCDK (approved):
Thanks!

Issue #440, opened 2023-01-03
Trying to plot cell transitions for generic FGW problem


Marius1311:
Hey there! I'm dealing with a generic FGW problem that I'm applying in the temporal domain. Naively, I was trying to plot the usual cell-transition heatmap using

```python
ct_desc = fgwp.cell_transition(source=8, 
                             target=14, 
                             source_groups={"leiden":all_clusters}, 
                             target_groups={"leiden":all_clusters}, 
                             forward=True)
mpl.cell_transition(fgwp, dpi=100, fontsize=10, return_fig=False)
```

I was expecting this either to work, or to tell me that this feature is not implemented for a generic FGW problem. Instread, I get
```
File /lustre/groups/ml01/code/marius.lange/moscot/src/moscot/problems/base/_mixins.py:435, in AnalysisMixin._annotation_aggregation_transition(self, source, target, annotation_key, annotations_1, annotations_2, df, tm, forward)
    433 func = self.push if forward else self.pull
    434 for subset in annotations_1:
--> 435     result = func(  # TODO(@MUCDK) check how to make compatible with all policies
    436         start=source,
    437         end=target,
    438         data=annotation_key,
    439         subset=subset,
    440         normalize=True,
    441         return_all=False,
    442         scale_by_marginals=False,
    443         split_mass=False,
    444         key_added=None,
    445         return_data=True,
    446     )
    447     df["distribution"] = result
    448     cell_dist = df[df[annotation_key].isin(annotations_2)].groupby(annotation_key).sum()

TypeError: push() missing 2 required positional arguments: 'source' and 'target'
```

If this plotting function is not meant to be used with a generic problem class, I think it would be nice to raise some sort of not implemented error. The exact same command with the temporal problem yields the desired output.

MUCDK:
Hi @Marius1311 ,

Thanks for the great catch!

I think this is not due to `moscot.plotting.cell_transition', but rather to the method.

Hope that #443 resolves this.

PR #443, opened 2023-01-04
make push/pull always use source/target

MUCDK:


MUCDK:
resolved #440

src/moscot/problems/time/_mixins.py
```diff
@@ -740,23 +740,23 @@ def _interpolate_gex_randomly(
...
                 f"Expected interpolation parameter to be in interval `(0, 1)`, found `{interpolation_parameter}`."
             )
 
-        if start < intermediate < end:
-            return (intermediate - start) / (end - start)
+        if source < intermediate < target:
+            return (intermediate - source) / (target - source)
         raise ValueError(
-            f"Expected intermediate time point to be in interval `({start}, {end})`, found `{intermediate}`."
+            f"Expected intermediate time point to be in interval `({source}, {target})`, found `{intermediate}`."
```
michalk8, line 759: Can be `{source, target}`

src/moscot/plotting/_utils.py
```diff
@@ -359,8 +359,8 @@ def _plot_temporal(
     adata: AnnData,
     temporal_key: str,
     key_stored: str,
-    start: float,
-    end: float,
+    source: float,
```
michalk8, line 362: Think `source/target` should be `Any`.

michalk8 (changes requested):

src/moscot/plotting/_utils.py
MUCDK: In this case actually `float` because we expect `adata.obs[time_key]` to be of type `float`

src/moscot/problems/time/_mixins.py
michalk8: It will be still formatted the same, though slightly more efficient, but I agree, it's more confusing, let's keep as it is.

michalk8 (approved):

PR #450, opened 2023-01-23
make push/pull plot in right order

MUCDK:
This way the cells from the initial distribution will be always plotted on top.

### DIFF
--- a/src/moscot/plotting/_utils.py
+++ b/src/moscot/plotting/_utils.py
@@ -408,7 +408,7 @@ def _plot_temporal(
         else:
             titles = [f"{categories} at time {source if push else target} and {name}"]
     for i, ax in enumerate(axs):
-        with RandomKeys(adata, n=1, where="obs") as keys:
+        with RandomKeys(adata, n=2, where="obs") as keys:
             if time_points is None:
                 if scale:
                     adata.obs[keys[0]] = (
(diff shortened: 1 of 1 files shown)


src/moscot/plotting/_utils.py
```diff
@@ -441,6 +441,13 @@ def _plot_temporal(
                         raise ValueError(f"Not exactly two categories, found `{column.cat.categories}`.")
                     kwargs["palette"] = {vmax: cont_cmap.reversed()(0), vmin: cont_cmap(0), st: na_color}
                     adata.obs[keys[0]] = column.values
+                    adata.obs[keys[1]] = list(range(adata.n_obs))
+                    cells_with_vmax = adata[adata.obs[keys[0]] == vmax].obs[keys[1]].values
+                    cells_with_vmin = adata[adata.obs[keys[0]] == vmin].obs[keys[1]].values
+                    cells_with_st = adata[adata.obs[keys[0]] == st].obs[keys[1]].values
+                    indices = list(cells_with_st) + list(cells_with_vmin) + list(cells_with_vmax)
+                    adata = adata[indices, :]
```
MUCDK, line 449: @michalk8 we currently need to do a copy of the `AnnData` instance as the colors corresponding to `adata.obs[keys[0]]` have to be generated and then saved to `AnnData`. This only happens in the `sc.pl.embedding()` below.

What do you think is best to do? Maybe we could explicitly create the `adata.uns` entry? or just accept that we have to do a copy?

MUCDK:
replaced by #453

PR #469, opened 2023-02-13
Move ```_validate_annotations``` into ```_utils.py```

### LINKED ISSUE #468: ```_validate_annotations``` in ```AnalysisMixinProtocol``` can be moved into ```_utils.py```
```_validate_annotations``` does not need to be a class function of ```AnalysisMixinProtocol``` since it does not depend on it. Thus, I believe this should be moved into ```_utils.py```.

lucaeyring:
Fixes #468.


giovp:
@lucaeyring are you sure you are using latest ott-jax? also can you rebase from main to rerun tests? thanks!

MUCDK:
replaced by #506

PR #471, opened 2023-02-13
Feature/sparsify transport matrices

MUCDK:


giovp:
for 3.8 and docs ci need to pin `scanpy>1.9.x` in moscot, that should do the job.

src/moscot/_utils.py
```diff
@@ -1,20 +1,56 @@
...
+def require_solution(
+    wrapped: Callable[[Any], Any], instance: "BaseProblem", args: Tuple[Any, ...], kwargs: Mapping[str, Any]
+) -> Any:
+    """Check whether problem has been solved."""
+    from moscot.problems.base import OTProblem, BaseCompoundProblem  # type: ignore[attr-defined]
+
+    if isinstance(instance, OTProblem) and instance.solution is None:
+        raise RuntimeError("Run `.solve()` first.")
+    if isinstance(instance, BaseCompoundProblem) and instance.solutions is None:
+        raise RuntimeError("Run `.solve()` first.")
```
giovp, line 38:
```suggestion
    if isinstance(instance, (OTProblem, BaseCompoundProblem)) and instance.solution is None:
        raise RuntimeError("Run `.solve()` first.")
```

giovp: this should also work

src/moscot/_utils.py
```diff
@@ -1,20 +1,56 @@
...
 
 jit_kwargs = {"nogil": True, "cache": True, "fastmath": True}
 
+__all__ = ["require_prepare", "require_solution", "parallelize"]
+
+
+# TODO(michalk8): refactor using stage
+@wrapt.decorator
+def require_solution(
+    wrapped: Callable[[Any], Any], instance: "BaseProblem", args: Tuple[Any, ...], kwargs: Mapping[str, Any]
```
giovp, line 30: maybe rename `instance` to `problem`?

src/moscot/_utils.py
```diff
@@ -128,7 +164,7 @@ def wrapper(*args, **kwargs):
             collections = [collection[[ix], :] for ix in range(collection.shape[0])]  # type: ignore
         else:
             step = collection.shape[0] // n_split  # type: ignore[union-attr]
-            ixs = [np.arange(i * step, min((i + 1) * step, collection.shape[0])) for i in range(n_split)]  # type: ignore  # noqa:501
+            ixs = [np.arange(i * step, min((i + 1) * step, collection.shape[0])) for i in range(n_split)]  # type:ignore
             ixs[-1] = np.append(ixs[-1], np.arange(ixs[-1][-1] + 1, collection.shape[0]))  # type: ignore
```
giovp, line 168: what are the type ignore heres? do you mind adding the arg of the type ignore for clarity?

src/moscot/problems/generic/_generic.py
```diff
@@ -62,7 +62,7 @@ def prepare(
         --------
         %(ex_prepare)s
         """
-        self.batch_key = key
+        self.batch_key = key  # type:ignore[misc]
```
giovp, line 65: there are a lot of mypy type ignore it's a bit worrying tbh especially because they are scattered around the repo. Any idea on why would it be the case?

src/moscot/solvers/_output.py
```diff
@@ -167,6 +177,73 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+        mode: Literal["threshold", "percentile", "min_1"] = SparsifyMode.THRESHOLD,
+        threshold: float = 1e-8,
+        batch_size: int = 1024,
+        n_samples: Optional[int] = None,
+    ) -> None:
+        """
+        Sparsify the transport matrix.
+
+        This function sets all entries of the transport matrix below `threshold` to 0 and
+        returns the result as a :class:`scipuy.sparse.csr_matrix`.
```
giovp, line 192:
```suggestion
        returns the result as a :class:`scipy.sparse.csr_matrix`.
```

src/moscot/solvers/_output.py
```diff
@@ -167,6 +177,73 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+        We encourage users not to use this function unless it is necessary. Note that this
+        does not necessarily reduce memory complexity.
```
giovp, line 194: maybe this could go under `Note` ?

src/moscot/solvers/_output.py
```diff
@@ -167,6 +177,73 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+
+        Parameters
+        ----------
+        mode
+            Which threshold to use for sparsification. Can be one of
+            - "threshold" - threshold below which entries are set to 0.0.
+            - "percentile" - determine threshold by percentile below which entries are set to 0. Hence, between 0 and
+            100.
+            - "min_1" - choose the threshold such that each row has at least one non-zero entry.
```
giovp, line 203:
```suggestion
            Which threshold to use for sparsification. Can be one of:

            	- "threshold" - threshold below which entries are set to 0.0.
            	- "percentile" - determine threshold by percentile below which entries are set to 0. Hence, between 0 and 100.
            	- "min_1" - choose the threshold such that each row has at least one non-zero entry.

```

giovp: think without spaces sphinx would throw error.

giovp (changes requested):
minor things, only thing a bit concerning are all the mypy ignores introduces across many modules, is that due to the decorator? In particular, the ones with `[misc]` are a bit worrying.

src/moscot/_utils.py
MUCDK: can't see why tbh. once we check for `instance.solution` once for `instance.solutions`

giovp: ahh good points missed the `solution`/`solutions` apologies

src/moscot/problems/generic/_generic.py
MUCDK: resolved it by downgrading mypy again

src/moscot/solvers/_output.py
```diff
@@ -1,16 +1,22 @@
 from abc import ABC, abstractmethod
 from copy import copy
-from typing import Any, Tuple, Callable, Iterable, Optional
+from typing import Any, List, Tuple, Literal, Callable, Iterable, Optional
 from functools import partial
 
+from scipy.sparse import hstack as sparse_hstack, csr_matrix
```
michalk8, line 6: Think it's better to use `import scipy.sparse as sp`

src/moscot/solvers/_output.py
```diff
@@ -167,6 +177,79 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+    @d.dedent
+    def compute_sparsification(
+        self,
+        mode: Literal["threshold", "percentile", "min_1"] = SparsifyMode.THRESHOLD,
+        threshold: float = 1e-8,
+        batch_size: int = 1024,
+        n_samples: Optional[int] = None,
+    ) -> None:
+        """
+        Sparsify the transport matrix.
```
michalk8, line 189: Sparsify the :attr:`transport_matrix`.

src/moscot/solvers/_output.py
```diff
@@ -167,6 +177,79 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+            x = np.eye(self.shape[1], max(n_samples, self.shape[1]))
+            res = self.pull(x, scale_by_marginals=False)  # tmap @ indicator_vectors
+            thr = np.percentile(res, threshold)
+        elif mode == SparsifyMode.MIN_1:
+            thr = np.inf
+            for batch in range(0, self.shape[1], batch_size):
+                x = np.eye(self.shape[1], min(batch_size, self.shape[1] - batch), -(min(batch, self.shape[1])))
+                res = self.pull(x, scale_by_marginals=False)  # tmap @ indicator_vectors
+                thr_batch = res.max(axis=1).min()
+                thr = thr_batch if thr_batch < thr else thr
```
michalk8, line 236: Can be simplified as `thr = min(thr, res.max(axis=1).min())`

src/moscot/solvers/_output.py
```diff
@@ -167,6 +177,79 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+            thr -= 1e-12  # necessary for the edge case
+        else:
+            raise NotImplementedError(mode)
+
+        tmaps_sparse: List[csr_matrix] = []
+        for batch in range(0, self.shape[1], batch_size):
```
michalk8, line 242: Small issue: if n << m, this is fairly inefficient. Would consider doing push in that case (and iterating over rows), since matmat operations are really efficient (and looping isn't).

src/moscot/solvers/_output.py
```diff
@@ -167,6 +177,79 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+            x = np.eye(self.shape[1], min(batch_size, self.shape[1] - batch), -(min(batch, self.shape[1])))
+            res = self.pull(x, scale_by_marginals=False)  # tmap @ indicator_vectors
+            res[res <= thr] = 0
+            tmaps_sparse.append(csr_matrix(res))
+        self._sparsified_tmap = sparse_hstack(tmaps_sparse)
+
+    @property
+    def sparsified_tmap(self) -> Optional[csr_matrix]:
```
michalk8, line 250: From API standpoint: do you really want to save the sparse matrix in the `output`?
It's not being used in any of the downstream functions like BaseOutput.{push,pull}...
As an alternative, `compute_sparsification` could just return the sparse matrix or even return `MatrixSolverOutput` instead so that user can use `push/pull`, etc. For the latter feature, #492 is needed, aas `MatrixSolverOutput` is no longer abstract.

src/moscot/solvers/_output.py
```diff
@@ -167,6 +177,79 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+            res = self.pull(x, scale_by_marginals=False)  # tmap @ indicator_vectors
+            thr = np.percentile(res, threshold)
+        elif mode == SparsifyMode.MIN_1:
+            thr = np.inf
+            for batch in range(0, self.shape[1], batch_size):
+                x = np.eye(self.shape[1], min(batch_size, self.shape[1] - batch), -(min(batch, self.shape[1])))
+                res = self.pull(x, scale_by_marginals=False)  # tmap @ indicator_vectors
+                thr_batch = res.max(axis=1).min()
+                thr = thr_batch if thr_batch < thr else thr
+            thr -= 1e-12  # necessary for the edge case
```
michalk8, line 237: Wouldn't changing the comparison down the line from `<=` to `<` also achieve the same?

tests/solvers/test_base_solver.py
```diff
@@ -0,0 +1,65 @@
...
+            mode="percentile", threshold=threshold, batch_size=batch_size, n_samples=int(0.5 * shape[0])
+        )
+        res = output.sparsified_tmap
+        assert isinstance(res, csr_matrix)
+        assert res.shape == shape
+        assert np.all(res.data >= 0.0)
+        if threshold == 0:
+            np.testing.assert_allclose(res.A, tmap, rtol=RTOL, atol=ATOL)
+        if threshold == 100:
+            assert np.all(res.data == 0.0)
```
michalk8, line 65: Use `assert res.nnz == 0`

src/moscot/solvers/_output.py
```diff
@@ -167,6 +177,79 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+
+        Parameters
+        ----------
+        mode
+            Which threshold to use for sparsification. Can be one of:
+
+                - "threshold" - threshold below which entries are set to 0.0.
+                - "percentile" - determine threshold by percentile below which entries are set to 0. Hence, between 0
+                  and 100.
+                - "min_1" - choose the threshold such that each row has at least one non-zero entry.
```
michalk8, line 202: Maybe call this `min_row`?

tests/solvers/test_base_solver.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class TestBaseSolverOutput:
+    @pytest.mark.parametrize("batch_size", [1, 4])
+    @pytest.mark.parametrize("threshold", [0.0, 1e-1, 1.0])
+    @pytest.mark.parametrize("shape", [(7, 2), (91, 103)])
+    def test_compute_sparsification_threshold(self, batch_size: int, threshold: float, shape: Tuple[int, int]) -> None:
+        rng = np.random.RandomState(42)
+        tmap = np.abs(rng.rand(shape[0], shape[1]))
+        tmap = tmap / tmap.sum()
```
michalk8, line 18: Why are you normalizing here? If not necessary, would remove.

tests/solvers/test_base_solver.py
```diff
@@ -0,0 +1,65 @@
...
+        output = MockSolverOutput(tmap)
+        output.compute_sparsification(mode="threshold", threshold=threshold, batch_size=batch_size)
+        res = output.sparsified_tmap
+        assert isinstance(res, csr_matrix)
+        assert res.shape == shape
+        assert np.all(res.data >= 0)
+        if threshold == 0.0:
+            np.testing.assert_allclose(res.A, tmap, rtol=RTOL, atol=ATOL)
+        if threshold == 1e-8:
```
michalk8, line 27: elif

tests/solvers/test_base_solver.py
MUCDK: To make it a bit more similar to a transport matrix, but can also remove

src/moscot/solvers/_output.py
MUCDK: I see your point and was thinking about the same. 
At the moment I save it as an attribute for the user to have more control while making it easily accessible by CellRank.

I.e., if I want to use the CR kernel, I can first create sparsification and then CellRank reads the sparsified transport matrix from the moscot class. If we only call sparsification after instantiating the moscot-CR kernel, the user cannot check the sparsified transport matrix before construction of the CR kernel.

I.e., we want CR to only read out results from moscot, rather than calling functions within moscot.

But open for discussions here

src/moscot/solvers/_output.py
```diff
@@ -167,6 +177,79 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+
+        threshold
+            Threshold or percentile depending on `mode`. If `mode` is `min_1`, `threshold` can be neglected.
+        batch_size
+            How many rows of the transport matrix to sparsify per batch.
+        n_samples
+            If `mode` is `percentile`, determine the number of samples based on which the percentile
+            is computed stochastically. Note this means that a matrix of shape `[n_samples, transport_matrix.shape[1]]`
+            has to be instantiated. If `None`, `n_samples` is set to batch_size.
```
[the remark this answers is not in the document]
MUCDK: I made all arguments but `n_samples` without Optional value, but `n_samples` must be optional as it is only needed if `mode=="percentile"`. Alternatively, could throw error when we have `mode=="percentile"` and `n_samples` is `None`.

src/moscot/solvers/_output.py
MUCDK: For the sake of reducing complexity I would prefer to keep it simple unless you have a strong opinion.

src/moscot/solvers/_output.py
michalk8: I don't think saving it as `csr` makes sense, since within `moscot`, you're not really using this (as explained in the `Returns` section).
I'd prefer this function to be more useful than from an interfacing standpoint to CR and would prefer to return `MatrixSolverOutput` rather than setting an attribute which is public and most importantly, not used anywhere, as then we/the user can easily quantify what effect it has on push/pull, etc.
From the interface standpoint, since it's no longer saved, we could pass kwargs to the `MongeKernel` that sparsifies this as and inside use `{k: v.sparsify(**kwargs) for k, prob in prob.solution.items()}`

src/moscot/base/output.py
```diff
@@ -168,6 +171,93 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+            res = self.pull(x, scale_by_marginals=False)  # tmap @ indicator_vectors
+            res[res < thr] = 0
+            tmaps_sparse.append(sp.csr_matrix(res))
+        self._sparsified_tmap = sp.hstack(tmaps_sparse)
+
+    @property
+    def sparse_transport_matrix(self) -> Optional[sp.csr_matrix]:
+        """Sparsified transport matrix.
+
+        This is `None` unless :meth:`moscot.base.output.BaseSolverOutput.sparsify`
```
michalk8, line 256: Just :meth:\`sparsify\`

src/moscot/solvers/_output.py
michalk8: It shouldn't be too complex, e.g.,:
```python
n, m = self.shape
if n < m
   k = n
   func = self.push
else:
   k = m
   func = self.pull

for i in range(0, k, batch_size)
   x = np.eye(...)
   res = func(x, scale_by_marginals=False)

Can you please add this?

File: src/moscot/base/output.py [hunk 20]
```diff
@@ -168,6 +171,93 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+            Random seed needed for sampling if ``mode = 'percentile'``.
+
+        Returns
+        -------
+        Nothing, but adds the sparsified transport matrix to :attr:`self.sparsified_tmap`.
+
+        Note
+        ----
+        This function only serves for interfacing software which has to instantiate the transport matrix. Within
+        :mod:`moscot`, there is no point in using this function. We encourage users not to use this function unless it
```
michalk8, line 216: I think this comment can scare the user, esp. the last sentence. Would soften this a bit.

src/moscot/base/output.py
```diff
@@ -168,6 +171,93 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+        n_samples
+            If ``mode = 'percentile'``, determine the number of samples based on which the percentile
+            is computed stochastically. Note this means that a matrix of shape `[n_samples, transport_matrix.shape[1]]`
+            has to be instantiated. If `None`, ``n_samples`` is set to batch_size.
+        seed
+            Random seed needed for sampling if ``mode = 'percentile'``.
+
+        Returns
+        -------
+        Nothing, but adds the sparsified transport matrix to :attr:`self.sparsified_tmap`.
```
michalk8, line 211: :attr:\`sparsified_tmap\`

src/moscot/base/output.py
```diff
@@ -168,6 +171,93 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+    def sparsify(
+        self,
+        mode: Literal["threshold", "percentile", "min_row"],
+        threshold: Optional[float] = None,
+        batch_size: int = 1024,
+        n_samples: Optional[int] = None,
+        seed: Optional[int] = None,
+    ) -> None:
+        """
+        Sparsify the :attr:transport_matrix.
```
michalk8, line 183: :attr:\`transport_matrix\`

src/moscot/base/output.py
```diff
@@ -168,6 +171,93 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+        mode
+            Which threshold to use for sparsification. Can be one of:
+
+                - 'threshold' - threshold below which entries are set to 0.0.
+                - 'percentile' - determine threshold by percentile below which entries are set to 0. Hence, between 0
+                  and 100.
+                - 'min_row' - choose the threshold such that each row has at least one non-zero entry.
+
+        threshold
+            Threshold or percentile depending on ``mode```. If ``mode = 'min_row'``, ``threshold``` can be neglected.
```
michalk8, line 199: Extra \` in the mode.

michalk8: Also in threshold.

src/moscot/base/output.py
```diff
@@ -168,6 +171,93 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+        """
+        Sparsify the :attr:transport_matrix.
+
+        This function sets all entries of the transport matrix below `threshold` to 0 and
+        returns the result as a :class:`~scipy.sparse.csr_matrix`.
+
+        Parameters
+        ----------
+        mode
+            Which threshold to use for sparsification. Can be one of:
```
michalk8, line 191: How to determine the value below which entries are set to 0:
  - \`'threshold'\` - explicitly passed using the \`\`threshold\`\` parameter.
  - \`'percentile'`\ - ...
  - ...

src/moscot/base/output.py
```diff
@@ -168,6 +171,93 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+        threshold: Optional[float] = None,
+        batch_size: int = 1024,
+        n_samples: Optional[int] = None,
+        seed: Optional[int] = None,
+    ) -> None:
+        """
+        Sparsify the :attr:transport_matrix.
+
+        This function sets all entries of the transport matrix below `threshold` to 0 and
+        returns the result as a :class:`~scipy.sparse.csr_matrix`.
```
michalk8, line 186: You don't need to mention the type in the docstring, it's already obvious from the return type.

tests/solvers/test_base_solver.py
```diff
@@ -0,0 +1,62 @@
...
+    @pytest.mark.parametrize("batch_size", [1, 4])
+    @pytest.mark.parametrize("threshold", [0.0, 1e-1, 1.0])
+    @pytest.mark.parametrize("shape", [(7, 2), (91, 103)])
+    def test_sparsify_threshold(self, batch_size: int, threshold: float, shape: Tuple[int, int]) -> None:
+        rng = np.random.RandomState(42)
+        tmap = np.abs(rng.rand(shape[0], shape[1]))
+        output = MockSolverOutput(tmap)
+        output.sparsify(mode="threshold", threshold=threshold, batch_size=batch_size)
+        res = output.sparse_transport_matrix
+        assert isinstance(res, csr_matrix)
```
michalk8, line 21: use `import scipy.sparse as sp` and `assert sp.isspmatrix_csr`

src/moscot/base/output.py
michalk8: Wrong attribute link, should be: sparse_transport_matrix
But would prefer this to return `MatrixSolverOutput` anyways.

tests/solvers/test_base_solver.py
```diff
@@ -0,0 +1,62 @@
...
+        assert res.shape == shape
+        assert np.all(res.data >= 0)
```
michalk8, line 23: Use `np.testing`

tests/solvers/test_base_solver.py
```diff
@@ -0,0 +1,62 @@
...
+    @pytest.mark.parametrize("shape", [(7, 2), (91, 103)])
+    def test_sparsify_minrow(self, batch_size: int, shape: Tuple[int, int]) -> None:
+        rng = np.random.RandomState(42)
+        tmap = np.abs(rng.rand(shape[0], shape[1])) + 1e-3  # make sure it's not 0
+        output = MockSolverOutput(tmap)
+        output.sparsify(mode="min_row", batch_size=batch_size)
+        res = output.sparse_transport_matrix
+        assert isinstance(res, csr_matrix)
+        assert res.shape == shape
+        assert np.all(res.data >= 0.0)
```
michalk8, line 44: Using np.testing

tests/solvers/test_base_solver.py
```diff
@@ -0,0 +1,62 @@
...
+    @pytest.mark.parametrize("shape", [(7, 2), (91, 103)])
+    def test_sparsify_percentile(self, batch_size: int, threshold: float, shape: Tuple[int, int]) -> None:
+        rng = np.random.RandomState(42)
+        tmap = np.abs(rng.rand(shape[0], shape[1])) + 1e-3  # make sure it's not 0
+        output = MockSolverOutput(tmap)
+        output.sparsify(mode="percentile", threshold=threshold, batch_size=batch_size, n_samples=shape[0], seed=42)
+        res = output.sparse_transport_matrix
+        assert isinstance(res, csr_matrix)
+        assert res.shape == shape
+        assert np.all(res.data >= 0.0)
```
michalk8, line 58: Use np.testing

tests/solvers/test_base_solver.py
michalk8: Actually, these checks aren't really necessary, would remove.

tests/solvers/test_base_solver.py
```diff
@@ -0,0 +1,62 @@
...
+        if threshold == 0:
+            assert np.sum(tmap != res.A) < res.shape[0] * res.shape[1] * 0.1  # this only holds with probability < 1
```
michalk8, line 60: Would extract `n, m = shape` and use it here

michalk8 (changes requested):

src/moscot/solvers/_output.py
MUCDK: okay, let's do it this way.

src/moscot/base/output.py
MUCDK: Now kept it as the return type is `MatrixSolverOutput`

tests/solvers/test_base_solver.py
MUCDK: Why are they not necessary? too obvious you mean?

michalk8: Yes, from the return type - you expect this is honored by how we specified the types so no need to check this.

MUCDK: btw just needed it when adapting the code as I initially transposed the `csr_matrix` to arrive at a `csc_matrix` (not doing this any more as I transpose before conversion to a `csr_matrix`.

src/moscot/base/output.py
MUCDK: slightly adapted the docs, so this does not apply any more

src/moscot/base/output.py
```diff
@@ -168,6 +171,98 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+        value: Optional[float] = None,
+        batch_size: int = 1024,
+        n_samples: Optional[int] = None,
+        seed: Optional[int] = None,
+    ) -> "MatrixSolverOutput":
+        """
+        Sparsify the :attr:`transport_matrix`.
+
+        This function sets all entries of the transport matrix below `threshold` to 0 and
+        returns an instance of :class:`moscot.base.MatrixSolverOutput` with the the
```
michalk8, line 186: I wouldn't link to the `MatrixSolverOutput`, it's already a in the type hint (not to mention this link isn't correct)>

src/moscot/base/output.py
```diff
@@ -168,6 +171,98 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+    def sparsify(
+        self,
+        mode: Literal["threshold", "percentile", "min_row"],
+        value: Optional[float] = None,
+        batch_size: int = 1024,
+        n_samples: Optional[int] = None,
+        seed: Optional[int] = None,
+    ) -> "MatrixSolverOutput":
+        """
+        Sparsify the :attr:`transport_matrix`.
```
michalk8, line 183: Should be on the first line (will unify the docstring style in #490, just to get used to it).

src/moscot/base/output.py
```diff
@@ -168,6 +171,98 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+        This function sets all entries of the transport matrix below `threshold` to 0 and
+        returns an instance of :class:`moscot.base.MatrixSolverOutput` with the the
+        sparsified transport matrix stored as a :class:`~scipy.sparse.csr_matrix`.
+
+        Parameters
+        ----------
+        mode
+            How to determine the value below which entries are set to 0:
+
+                - 'threshold' - threshold below which entries are set to 0.0.
```
michalk8, line 194: Remove the `.0` in `0.0` (also in other places).

src/moscot/base/output.py
```diff
@@ -168,6 +171,98 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+            Value to use for sparsification depending on ``mode``:
+
+                - `threshold` - `value` sets the threshold below which entries are set to 0.0.
+                - `percentile` - `value` is the percentile below which entries are set to 0.0.
+                - `min_row` - `value` is not used.
+        batch_size
+            How many rows of the transport matrix to sparsify per batch.
+        n_samples
+            If ``mode = 'percentile'``, determine the number of samples based on which the percentile
+            is computed stochastically. Note this means that a matrix of shape `[n_samples, transport_matrix.shape[1]]`
```
michalk8, line 209: This is no longer accurate, try:

... is computed stochastically by instantiating a matrix of shape `min(n_samples, *self.shape)`.

src/moscot/base/output.py
```diff
@@ -168,6 +171,98 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+            has to be instantiated. If `None`, ``n_samples`` is set to ``batch_size``.
+        seed
+            Random seed needed for sampling if ``mode = 'percentile'``.
+
+        Returns
+        -------
+        Returns :class:`moscot.base.MatrixSolverOutput` with a sparsified transport matrix.
```
michalk8, line 216: Again, don't mention the type, would write just: The sparsified transport matrix.

src/moscot/base/output.py
```diff
@@ -168,6 +171,98 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+
+        Note
```
michalk8, line 218: Suggestion: instead of this note (which is imho fairly important), add a warning after the description as
```rst

... stored as a :class:`~scipy.sparse.csr_matrix`.

..  warning::

    This function server only for interfacing ...

```

src/moscot/base/output.py
```diff
@@ -168,6 +171,98 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+            thr = np.inf
+            for batch in range(0, m, batch_size):
+                x = np.eye(m, min(batch_size, m - batch), -(min(batch, m)))
+                res = self.pull(x, scale_by_marginals=False)  # tmap @ indicator_vectors
+                thr = min(thr, res.max(axis=1).min())
+        else:
+            raise NotImplementedError(mode)
+
+        if n < m:
+            k = n
```
michalk8, line 249: All these assignment can be done in 1 line to save space (also in the else branch).

src/moscot/base/output.py
```diff
@@ -168,6 +171,98 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+        else:
+            k = m
+            func = self.pull
+            fn_stack = sp.hstack
+        tmaps_sparse: List[sp.csr_matrix] = []
+        for batch in range(0, k, batch_size):
+            x = np.eye(k, min(batch_size, k - batch), -(min(batch, k)))
+            res = func(x, scale_by_marginals=False)
+            res[res < thr] = 0
+            tmaps_sparse.append(sp.csr_matrix(res.T if func == self.push else res))
```
michalk8, line 261: Don't use `func == self.push`, use `func is self.push`

tests/solvers/test_base_solver.py
```diff
@@ -0,0 +1,67 @@
...
+class TestBaseSolverOutput:
+    @pytest.mark.parametrize("batch_size", [1, 4])
+    @pytest.mark.parametrize("threshold", [0.0, 1e-1, 1.0])
+    @pytest.mark.parametrize("shape", [(7, 2), (91, 103)])
+    def test_sparsify_threshold(self, batch_size: int, threshold: float, shape: Tuple[int, int]) -> None:
+        rng = np.random.RandomState(42)
+        tmap = np.abs(rng.rand(shape[0], shape[1]))
+        output = MockSolverOutput(tmap)
+        mso = output.sparsify(mode="threshold", value=threshold, batch_size=batch_size)
+        assert isinstance(mso, MatrixSolverOutput)
```
michalk8, line 21: This check is not redundant, as I was explaining, since you're accessing it's attributes later (and if it weren't of the correct type, it would fail).

tests/solvers/test_base_solver.py
```diff
@@ -0,0 +1,67 @@
...
+    def test_sparsify_minrow(self, batch_size: int, shape: Tuple[int, int]) -> None:
+        rng = np.random.RandomState(42)
+        tmap = np.abs(rng.rand(shape[0], shape[1])) + 1e-3  # make sure it's not 0
+        output = MockSolverOutput(tmap)
+        mso = output.sparsify(mode="min_row", batch_size=batch_size)
+        assert isinstance(mso, MatrixSolverOutput)
+        res = mso.transport_matrix
+        assert isinstance(res, sp.csr_matrix)
+        assert res.shape == shape
+        np.testing.assert_array_equal(res.data >= 0.0, np.full(res.data.shape, True))
```
michalk8, line 47: You don't have to run `np.full(res.data.shape)`, simply `True` is ok, since it will be broadcasted (change also in other places).

src/moscot/base/output.py
```diff
@@ -168,6 +171,98 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+
+                - 'threshold' - threshold below which entries are set to 0.0.
+                - 'percentile' - determine threshold by percentile below which entries are set to 0.0. Hence, between 0
+                  and 100.
+                - 'min_row' - choose the threshold such that each row has at least one non-zero entry.
+
+        value
+            Value to use for sparsification depending on ``mode``:
+
+                - `threshold` - `value` sets the threshold below which entries are set to 0.0.
```
michalk8, line 202: Can you also add the `'`, as \`'threshold'\`? I wanna signify this is a string, not some internal constant.

michalk8 (changes requested):

src/moscot/base/output.py
MUCDK: Do you mean completely dropping the piece of information that we return an instance of `MatrixSolverOutput`? That's difficult I think because it's useful to describe that we save the `transport_matrix` as a `csr_matrix`. For now, I just remove the link.

src/moscot/base/output.py
michalk8: Ok, but don't mention the type again, it's in the annotation.

src/moscot/base/output.py
michalk8: Only the former is good.

src/moscot/base/output.py
```diff
@@ -245,20 +244,13 @@ def sparsify(
...
-            func = self.pull
-            fn_stack = sp.hstack
+        k, func, fn_stack = (n, self.push, sp.vstack) if n < m else (m, self.pull, sp.hstack)
         tmaps_sparse: List[sp.csr_matrix] = []
         for batch in range(0, k, batch_size):
             x = np.eye(k, min(batch_size, k - batch), -(min(batch, k)))
             res = func(x, scale_by_marginals=False)
             res[res < thr] = 0
-            tmaps_sparse.append(sp.csr_matrix(res.T if func == self.push else res))
+            tmaps_sparse.append(sp.csr_matrix(res.T if func is self.push else res))
```
MUCDK, line 253: @michalk8 this does not work, so will revert back to `==`

src/moscot/base/output.py
MUCDK: not sure if I understand what you mean , but please have a look at the changes.

michalk8 (changes requested):

src/moscot/base/output.py
```diff
@@ -167,6 +170,90 @@ def chain(self, outputs: Iterable["BaseSolverOutput"], scale_by_marginals: bool
...
+            n_samples = n_samples if n_samples is not None else batch_size
+            k = min(n_samples, n)
+            x = np.zeros((m, k))
+            rows = rng.choice(m, size=k)
+            x[rows, np.arange(k)] = 1.0
+            res = self.pull(x, scale_by_marginals=False)  # tmap @ indicator_vectors
+            thr = np.percentile(res, value)
+        elif mode == "min_row":
+            thr = np.inf
+            for batch in range(0, m, batch_size):
```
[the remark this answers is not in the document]
MUCDK: you mean we should also do it for the stochastic determination of the percentile?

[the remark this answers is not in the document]
michalk8: I mean for the `min_row`.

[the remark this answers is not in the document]
MUCDK: Just realized that we cannot do it here, because we always look for each *row* to have at least one non-zero element. Hence, we need to instantiate each row completely (and not alternatively each row OR each column).

Agree?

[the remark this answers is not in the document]
michalk8: Good point, let's keep it as it is.

src/moscot/base/output.py
michalk8: This should work, maybe you tried with `autoreload`?

michalk8:
@MUCDK https://github.com/theislab/moscot/pull/471/commits/03274dc720a5acd609e4f642b2845a5dab634570 broke the linter, can you pls undo all the doc-related changes?

src/moscot/base/output.py
MUCDK: I tried it again, and locally tests fail again. I just now pushed another version with `is`, let's see whether CI fails.

michalk8: It might be because `push` is decorated... Maybe `if n < m` is better in this case.

michalk8 (approved):
LGTM, thanks @MUCDK !

PR #472, opened 2023-02-17
quick fix for cell transition in dummy policy case

giovp:


src/moscot/problems/base/_compound_problem.py
```diff
@@ -305,7 +305,7 @@ def _(
         _ = kwargs.pop("target", None)  # make compatible with Explicit/Ordered policy
         for src, tgt in self._policy.plan(
             explicit_steps=kwargs.pop("explicit_steps", None),
-            filter=source,  # type: ignore [arg-type]
+            filter=source if not isinstance(self._policy, DummyPolicy) else None,  # type: ignore [arg-type]
```
michalk8, line 308: Think it would be slightly cleaner override `_filter_plan` in `DummyPolicy` to be a no-op.

giovp:
close in favour of #474

PR #477, opened 2023-02-22
add normalization in alignment step

giovp:


giovp:
closed in favour of #478

PR #492, opened 2023-03-13
Feature/new structure

### LINKED ISSUE #483: Simplify/clean package structure

michalk8:
closes #483


giovp:
hi @michalk8 , I fear this PR will most likely break most of the analysis/benchmarks scripts we have. If that's the case (I feel it most probably will be), I would not merge this until manuscript is submitted.

michalk8:
> hi @michalk8 , I fear this PR will most likely break most of the analysis/benchmarks scripts we have. If that's the case (I feel it most probably will be), I would not merge this until manuscript is submitted.

From API standpoint, I just renamed `spatio_temporal` directory to `spatiotemporal` and maybe some other things that I'm forgetting.

giovp:
> maybe some other things that I'm forgetting.

this is exactly what I'm worried about and why I would pause this for the moment 😅

MUCDK:
> > maybe some other things that I'm forgetting.
> 
> this is exactly what I'm worried about and why I would pause this for the moment 😅

As discussed, I think this should be fine and not change anything in our scripts.

src/moscot/_constants/_constants.py
MUCDK, line 1: What's the idea behind deleting all enum classes? Is it for documentation reasons?

MUCDK: i.e. readability

src/moscot/backends/ott/_utils.py
```diff
@@ -1,18 +1,20 @@
...
 from moscot._logging import logger
+from moscot._types import ArrayLike, ScaleCost_t
 
 
 def _compute_sinkhorn_divergence(
     point_cloud_1: ArrayLike,
     point_cloud_2: ArrayLike,
     a: Optional[ArrayLike] = None,
     b: Optional[ArrayLike] = None,
+    # TODO(MUCDK): why is this so high?
```
MUCDK, line 17: Nice spot. When I wrote this function I used it for comparing distances between distributions of non-scaled gene expression data, which sometimes covers a big range, as it is not normalized (and we don't want to scale the cost because then different scales of differences are not recoverable any more).

I just want to make sure that it converges. Still, I think it is unreasonably high. Maybe set it to 1e-1?

src/moscot/base/cost.py
```diff
@@ -0,0 +1,70 @@
...
+    Parameters
+    ----------
+    adata
+        Annotated data object.
+    attr
+        Attribute of :class:`anndata.AnnData` used when computing the cost.
+    key
+        Key in the ``attr`` of :class:`anndata.AnnData` used when computing the cost.
+    dist_key
+        Helper key which determines into which subset ``adata`` belongs.
```
MUCDK, line 26: maybe 
`Helper key which determines which distribution ``adata`` belongs to.` ? 
or
`Identifier defining the key of the distribution ``adata`` belongs to.`?

src/moscot/problems/generic/_mixins.py
```diff
@@ -40,12 +43,13 @@ def cell_transition(
         target: K,
         source_groups: Optional[Str_Dict_t] = None,
         target_groups: Optional[Str_Dict_t] = None,
+        # TODO(MUCDK): unused arg
```
MUCDK, line 46: will handle this in a different PR, thanks.

src/moscot/problems/spatio_temporal/__init__.py
MUCDK, line 1: I think this is the only file which might cause our analysis scripts to fail (@zoepiran).

MUCDK (approved):
LGTM, maybe consider the minor comments regarding documentation.

src/moscot/_constants/_constants.py
michalk8: Yes, given the removal of `ModeEnum` (its implementation wasn't that great), it doesn't provide much benefit, not to mention the issues with typing that arose sometime.

src/moscot/backends/ott/_utils.py
michalk8: Ok, will use `1e-1`.

src/moscot/base/cost.py
michalk8: Thanks, used the former!

giovp (approved):

MUCDK:
I had to update the tests due to the temporal regression tests which, so far, had used the default parameters of `compute_sinkhorn_divergence`, which we changed here from 10 to 1e-1.

Moreover, the expected plots of the `cell_transition` are based on the generated "ground truth data", I regenerated this data for the abovementioned reason, hence I also had to update the expected plots.

michalk8:
Merging this, thanks @MUCDK @giovp !

Issue #476, opened 2023-02-20
`cell_transition` doesn't work when `source_groups != target_groups`


zoepiran:
calling `cell_transition` with different keys for `source_groups` and `target_groups` raises an error, see below. The same call with identical keys works. 
*works*:
```{python}
cell_transition = stp.cell_transition(
    source=start,
    target=end,
    source_groups="annotation",
    target_groups="annotation",
    forward=True
)
```
*error*:
```{python}
cell_transition = stp.cell_transition(
    source=start,
    target=end,
    source_groups="transition",
    target_groups="annotation",
    forward=True
)
```
-->
```{python}
/cs/labs/mornitzan/zoe.piran/research/projects/moscot/src/moscot/problems/time/_mixins.py in cell_transition(self, source, target, source_groups, target_groups, forward, aggregation_mode, batch_size, normalize, key_added)
    188         if TYPE_CHECKING:
    189             assert isinstance(self.temporal_key, str)
--> 190         return self._cell_transition(
    191             key=self.temporal_key,
    192             source=source,

/cs/labs/mornitzan/zoe.piran/research/projects/moscot/src/moscot/problems/base/_mixins.py in _cell_transition(self, source, target, source_groups, target_groups, key_added, **kwargs)
    139             **kwargs,
    140         )
--> 141         tm = self._cell_transition_online(
    142             source=source,
    143             target=target,

/cs/labs/mornitzan/zoe.piran/research/projects/moscot/src/moscot/problems/base/_mixins.py in _cell_transition_online(self, key, source, target, source_groups, target_groups, forward, aggregation_mode, other_key, other_adata, batch_size, normalize, **_)
    225             )
    226             if forward:
--> 227                 tm = self._annotation_aggregation_transition(
    228                     source=source,
    229                     target=target,

/cs/labs/mornitzan/zoe.piran/research/projects/moscot/src/moscot/problems/base/_mixins.py in _annotation_aggregation_transition(self, source, target, annotation_key, annotations_1, annotations_2, df, tm, forward)
    452             )
    453             df["distribution"] = result
--> 454             cell_dist = df[df[annotation_key].isin(annotations_2)].groupby(annotation_key).sum()
    455             cell_dist /= cell_dist.sum()
    456             tm.loc[subset, :] = [

/cs/labs/mornitzan/zoe.piran/venvzp/lib/python3.9/site-packages/pandas/core/frame.py in __getitem__(self, key)
   3805             if self.columns.nlevels > 1:
[...]

PR #502, opened 2023-03-23
fix bug in cell_transition and add test

MUCDK:


MUCDK:
handles #476

MUCDK:
@michalk8 any idea why linting is failing? 
It says the link does not exist although I checked, and it still works.

michalk8:
> @michalk8 any idea why linting is failing?

Yes, because some sites check if the request comes from a human, which in this case not. Please add the link to `linkcheck_ignore` in `docs/conf.py`.

src/moscot/base/problems/problem.py
```diff
@@ -90,9 +90,8 @@ def _split_mass(arr: ArrayLike) -> ArrayLike:
             if subset is None:  # allow for numeric values
                 data = np.asarray(adata.obs[data], dtype=float)
             else:
-                if isinstance(subset, str):
-                    subset = [subset]
-                data = np.asarray(adata.obs[data].isin(subset), dtype=float)
+                sset = subset if isinstance(subset, list) else [subset]  # type:ignore[list-item]
```
michalk8, line 93: I still slightly prefer the `isinstance(..., str)` check, as it's less constraining than the `list` check.

tests/problems/time/test_mixins.py
```diff
@@ -50,6 +50,36 @@ def test_cell_transition_full_pipeline(self, gt_temporal_adata: AnnData, forward
...
+        key_1 = config["key_1"]
+        key_2 = config["key_2"]
+        key_3 = config["key_3"]
+
+        gt_temporal_adata.obs["batch"] = gt_temporal_adata.obs["batch"].astype("category")
+        batches = set(gt_temporal_adata.obs["batch"].cat.categories)
+        problem = TemporalProblem(gt_temporal_adata)
+        problem = problem.prepare(key)
+        assert set(problem.problems.keys()) == {(key_1, key_2), (key_2, key_3)}
+        problem[(key_1, key_2)]._solution = MockSolverOutput(gt_temporal_adata.uns["tmap_10_105"])
```
michalk8, line 67: Pet peeve: no need for `()` here.

tests/problems/time/test_mixins.py
```diff
@@ -50,6 +50,36 @@ def test_cell_transition_full_pipeline(self, gt_temporal_adata: AnnData, forward
...
+
+        result = problem.cell_transition(
+            key_1,
+            key_2,
+            "cell_type",
+            "batch",
+            forward=forward,
+        )
+        assert isinstance(result, pd.DataFrame)
+        cell_types = set(gt_temporal_adata[gt_temporal_adata.obs[key] == key_1].obs["cell_type"].cat.categories)
```
michalk8, line 78: I wouldn't subset the adata, but directly the dataframe using `.loc`; this might work:
```python
mask = ...
cell_types = set(gt_temporal_adata.obs.loc[mask, "cell_type"].cat.categories)
batches = set(gt_temporal_adata.obs.loc[mask, "batch"].cat.categories)
```

michalk8 (changes requested):
Minor changes, otherwise looks good!

tests/problems/time/test_mixins.py
MUCDK: had to revert it as by only subsetting the data frame like this, the `.cat.categories` remain the same, and hence the tests fail.
Alternatively, I could reset the categories, but then I think this way it is easier.

tests/problems/time/test_mixins.py
```diff
@@ -50,6 +50,36 @@ def test_cell_transition_full_pipeline(self, gt_temporal_adata: AnnData, forward
...
+        batches = set(gt_temporal_adata[gt_temporal_adata.obs[key] == key_2].obs["batch"].cat.categories)
+        assert set(result.index) == cell_types
+        assert set(result.columns) == set(batches)
```
michalk8, line 81: Unnecessary `set`.

michalk8 (changes requested):
Minor issue; after it's fixed, please merge.

PR #511, opened 2023-04-03
fix warning in moscot costs

MUCDK:


src/moscot/base/cost.py
```diff
@@ -51,7 +51,7 @@ def __call__(self, *args: Any, **kwargs: Any) -> ArrayLike:
         The computed cost matrix.
         """
         cost = self._compute(*args, **kwargs)
-        if not np.all(np.isnan(cost)):
+        if not np.all(~np.isnan(cost)):
```
michalk8, line 54: `if np.any(np.isnan(cost))` is better.

michalk8 (changes requested):

Issue #409, opened 2022-11-14
`cell_transition` doesn't work for `MappingProblem`


giovp:
```
File ~/code/moscot/src/moscot/problems/base/_compound_problem.py:379, in BaseCompoundProblem.pull(self, *args, **kwargs)
    377 _ = kwargs.pop("return_data", None)
    378 _ = kwargs.pop("key_added", None)  # this should be handled by overriding method
--> 379 return self._apply(*args, forward=False, **kwargs)

File ~/code/moscot/src/moscot/problems/base/_utils.py:37, in attributedispatch.<locals>.wrapper(instance, *args, **kwargs)
     35 def wrapper(instance: Any, *args: Any, **kwargs: Any) -> Any:
     36     typ = type(getattr(instance, str(attr)))
---> 37     return dispatch(typ)(instance, *args, **kwargs)

File ~/code/moscot/src/moscot/problems/base/_compound_problem.py:272, in BaseCompoundProblem._(self, data, forward, scale_by_marginals, start, return_all, **kwargs)
    270 start = start if isinstance(start, list) else [start]
    271 _ = kwargs.pop("end", None)  # make compatible with Explicit/Ordered policy
--> 272 for src, tgt in self._policy.plan(
    273     explicit_steps=kwargs.pop("explicit_steps", None),
    274     filter=start,  # type: ignore [arg-type]
    275 ):
    276     problem = self.problems[src, tgt]
    277     fun = problem.push if forward else problem.pull

File ~/code/moscot/src/moscot/problems/_subset_policy.py:94, in SubsetPolicy.plan(self, filter, explicit_steps, **kwargs)
     92     plan = self._filter_plan(plan, filter=filter)
     93 if not len(plan):
---> 94     raise ValueError("Unable to create a plan, no steps were selected after filtering.")
     95 return plan

ValueError: Unable to create a plan, no steps were selected after filtering.
```

MUCDK:
@giovp this is fixed now, right?

MUCDK:
handled by #514

PR #514, opened 2023-04-04
enable `None` if aggregation mode is cell

MUCDK:


MUCDK:
closes #409

giovp:
this is really cool I believe this could fix the bug we discussed before

MUCDK:
> this is really cool I believe this could fix the bug we discussed before

That's the plan :)

michalk8 (approved):

PR #516, opened 2023-04-05
Feature/new datasets

michalk8:
@Marius1311 can you please update the docstrings?
Also, currently the figshare is private, so downloads won't work.


MUCDK (approved):
Can we remove TedSim then?

MUCDK:
> @Marius1311 can you please update the docstrings? Also, currently the figshare is private, so downloads won't work.

Why is it private? You do have access to moscot figshare, don't you? or do I have to make it public then?

michalk8:
> Can we remove TedSim then?

I don't know, will create an issues for this.

Merging this not to block @Marius1311 .

Marius1311:
> Can we remove TedSim then?

No, we're just adding additional datasets, we cannot remove TedSim.

PR #536, opened 2023-05-03
fix growth_rates and cell_costs

MUCDK:


src/moscot/problems/time/_mixins.py
```diff
@@ -449,20 +448,19 @@ def prior_growth_rates(self: TemporalMixinProtocol[K, B]) -> Optional[pd.DataFra
...
     def posterior_growth_rates(self: TemporalMixinProtocol[K, B]) -> Optional[pd.DataFrame]:
         """Return the posterior estimate of growth rates of the cells in the source distribution."""
-        # TODO(michalk8): FIXME
         cols = ["posterior_growth_rates"]
         df_list = [
             pd.DataFrame(problem.posterior_growth_rates, index=problem.adata.obs.index, columns=cols)
             for problem in self.problems.values()
         ]
-        tup = list(self)[-1]
+        indices_remaining = set(self.adata.obs_names) - {ind for df in df_list for ind in df.index}
```
michalk8, line 456: This can be done in a nicer way: I'd first concatenate the `df_list` and then use a 2nd concat to add the remaining indices.

src/moscot/problems/time/_mixins.py
```diff
@@ -427,20 +427,19 @@ def pull(
...
             pd.DataFrame(problem.prior_growth_rates, index=problem.adata.obs.index, columns=cols)
             for problem in self.problems.values()
         ]
-        tup = list(self)[-1]
+        indices_remaining = set(self.adata.obs_names) - {ind for df in df_list for ind in df.index}
         df_list.append(
             pd.DataFrame(
                 np.full(
-                    shape=(len(self.problems[tup].adata_tgt.obs), 1),
+                    shape=(len(indices_remaining), 1),
```
michalk8, line 439: For readability, `shape=len(indices_remaining), len(columns)`.
But I think `pandas` should be able to broadcast this:
```python
pd.DataFrame(np.nan, index=list(indices_remaining), columns=columns)
```

src/moscot/problems/time/_mixins.py
```diff
@@ -485,11 +483,11 @@ def cell_costs_source(self: TemporalMixinProtocol[K, B]) -> Optional[pd.DataFram
             )
             for problem in self.problems.values()
         ]
-        tup = list(self)[-1]
+        indices_remaining = set(self.adata.obs_names) - {ind for df in df_list for ind in df.index}
```
michalk8, line 486: Change the assertion above + use `self.solutions`.

michalk8 (changes requested):

src/moscot/plotting/_utils.py
```diff
@@ -489,12 +491,17 @@ def _plot_temporal(
...
                 **kwargs,
             )
     if suptitle is not None:
+        if fig is None:
+            raise ValueError("Cannot set `suptitle` of figure when `ax` is not `None`.")
         fig.suptitle(suptitle, fontsize=suptitle_fontsize)
     if save:
+        if fig is None:
+            raise ValueError("Figure cannot be saved when `ax` is not `None`.")
         fig.figure.savefig(save, bbox_inches="tight")
```
michalk8, line 505: I don't think you need `fig.figure` (this should hold `fig.figure is fig`).

pyproject.toml
```diff
@@ -47,7 +47,7 @@ maintainers = [
 dependencies = [
     "numpy>=1.20.0",
     "scipy>=1.7.0",
-    "pandas>=1.4.0",
+    "pandas<2.0",
```
michalk8, line 50: Please create an issue for this.

src/moscot/plotting/_utils.py
```diff
@@ -489,12 +491,17 @@ def _plot_temporal(
                 title=titles[i],
                 size=size,
                 ax=ax,
-                show=show,
+                show=False,
+                return_fig=False,
                 **kwargs,
             )
     if suptitle is not None:
+        if fig is None:
```
michalk8, line 499: I think we can extract the figure from the raveled axes, no?
Otherwise, I think error is too strong, would just warn.

src/moscot/problems/time/_mixins.py
```diff
@@ -427,102 +426,72 @@ def pull(
...
+        return pd.concat([df_1, df_2], verify_integrity=True)
 
     @property
     def posterior_growth_rates(self: TemporalMixinProtocol[K, B]) -> Optional[pd.DataFrame]:
         """Return the posterior estimate of growth rates of the cells in the source distribution."""
-        # TODO(michalk8): FIXME
         cols = ["posterior_growth_rates"]
         df_list = [
-            pd.DataFrame(problem.posterior_growth_rates, index=problem.adata.obs.index, columns=cols)
+            pd.DataFrame(problem.posterior_growth_rates, index=problem.adata.obs_names, columns=cols)
```
michalk8, line 445: Does this work when `posterior_growth_rates` are `None`?

src/moscot/problems/time/_mixins.py
```diff
@@ -427,102 +426,72 @@ def pull(
...
 
-    # TODO(michalk8): refactor me
     @property
     def cell_costs_source(self: TemporalMixinProtocol[K, B]) -> Optional[pd.DataFrame]:
         """Return the cost of a cell obtained by the potentials of the optimal transport solution."""
-        sol = list(self.problems.values())[0].solution
-        if TYPE_CHECKING:
-            assert isinstance(sol, BaseSolverOutput)
-        if sol.potentials is None:
+        cols = ["cell_cost_source"]
```
michalk8, line 457: `cols` should be defined after the 1st return

src/moscot/problems/time/_mixins.py
```diff
@@ -427,102 +426,72 @@ def pull(
...
+        sol = list(self.solutions.values())[0]
+        if sol is None or sol.potentials is None:
             return None
+
         df_list = [
             pd.DataFrame(
                 np.array(np.abs(problem.solution.potentials[0])),  # type: ignore[union-attr,index]
```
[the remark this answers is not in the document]
michalk8: All of this also should apply to `cell_costs_target` below.

michalk8 (changes requested):

src/moscot/plotting/_utils.py
MUCDK: No, I don't think so. Fine, will do a warning

src/moscot/problems/time/_mixins.py
MUCDK: Yes, it returns a `pd.DataFrame` with `NaN` entries. But I think you're right, let's just return `None` then.

src/moscot/problems/time/_mixins.py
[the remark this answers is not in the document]
michalk8: > Because it's about the absolute value of the derivative, not the direction, right?

Ok, can you then please add to the docs it returns the absolute values of the dual potentials so that it's clear?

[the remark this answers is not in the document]
MUCDK: Actually changed it now to returning not the `abs` value

michalk8 (approved):

PR #478, opened 2023-03-01
add spatial normalization

giovp:

PR #543, opened 2023-05-24
add spatial normalization

giovp:
same as #478

### DIFF
--- a/src/moscot/base/problems/compound_problem.py
+++ b/src/moscot/base/problems/compound_problem.py
@@ -118,8 +118,10 @@ class BaseCompoundProblem(BaseProblem, abc.ABC, Generic[K, B]):
             return {}
         if callback == "local-pca":
             callback = problem._local_pca_callback
-        if callback == "spatial-norm":
+        elif callback == "spatial-norm":
             callback = problem._spatial_norm_callback
+        else:
+            raise ValueError(f"Unknown callback `{callback!r}`.")
 
         if not callable(callback):
             raise TypeError("Callback is not a function.")
--- a/src/moscot/base/problems/compound_problem.py
+++ b/src/moscot/base/problems/compound_problem.py
@@ -118,10 +118,8 @@ class BaseCompoundProblem(BaseProblem, abc.ABC, Generic[K, B]):
             return {}
         if callback == "local-pca":
             callback = problem._local_pca_callback
-        elif callback == "spatial-norm":
+        if callback == "spatial-norm":
             callback = problem._spatial_norm_callback
-        else:
-            raise ValueError(f"Unknown callback `{callback!r}`.")
 
         if not callable(callback):
             raise TypeError("Callback is not a function.")
--- a/src/moscot/base/problems/compound_problem.py
+++ b/src/moscot/base/problems/compound_problem.py
@@ -118,6 +118,8 @@ class BaseCompoundProblem(BaseProblem, abc.ABC, Generic[K, B]):
             return {}
         if callback == "local-pca":
             callback = problem._local_pca_callback
+        if callback == "spatial-norm":
+            callback = problem._spatial_norm_callback
 
         if not callable(callback):
             raise TypeError("Callback is not a function.")
--- a/src/moscot/base/problems/problem.py
+++ b/src/moscot/base/problems/problem.py
@@ -510,10 +510,10 @@ class OTProblem(BaseProblem):
     def _spatial_norm_callback(
         term: Literal["x", "y"],
         adata: AnnData,
-        spatial_key: str,
         adata_y: Optional[AnnData] = None,
         **kwargs: Any,
     ) -> Dict[Literal["x", "y"], TaggedArray]:
+        spatial_key = kwargs["spatial_key"]
         if term == "x":
             spatial = adata.obsm[spatial_key]
         if term == "y":
@@ -521,6 +521,7 @@ class OTProblem(BaseProblem):
                 spatial = adata_y.obsm[spatial_key]
             else:
                 raise ValueError("When `term` is `y` `adata_y` cannot be `None`.")
+        logger.info(f"Normalizing spatial coordinates of `{term}`.")
         spatial = (spatial - spatial.mean()) / spatial.std()
         return {term: TaggedArray(spatial, tag=Tag.POINT_CLOUD)}
 
--- a/src/moscot/base/problems/problem.py
+++ b/src/moscot/base/problems/problem.py
@@ -517,10 +517,10 @@ class OTProblem(BaseProblem):
         if term == "x":
             spatial = adata.obsm[spatial_key]
         if term == "y":
-            if isinstance(adata_y, AnnData):
-                spatial = adata_y.obsm[spatial_key]
-            else:
+            if adata_y is None:
                 raise ValueError("When `term` is `y`, `adata_y` cannot be `None`.")
+            spatial = adata_y.obsm[spatial_key]
+
         logger.info(f"Normalizing spatial coordinates of `{term}`.")
         spatial = (spatial - spatial.mean()) / spatial.std()
         return {term: TaggedArray(spatial, tag=Tag.POINT_CLOUD)}
--- a/src/moscot/base/problems/problem.py
+++ b/src/moscot/base/problems/problem.py
@@ -506,6 +506,24 @@ class OTProblem(BaseProblem):
             return {term: TaggedArray(x, tag=Tag.POINT_CLOUD)}
         raise ValueError(f"Expected `term` to be one of `x`, `y`, or `xy`, found `{term!r}`.")
 
+    @staticmethod
+    def _spatial_norm_callback(
+        term: Literal["x", "y"],
+        adata: AnnData,
+        spatial_key: str,
+        adata_y: Optional[AnnData] = None,
+        **kwargs: Any,
+    ) -> Dict[Literal["x", "y"], TaggedArray]:
+        if term == "x":
+            spatial = adata.obsm[spatial_key]
+        if term == "y":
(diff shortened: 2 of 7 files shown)


src/moscot/problems/space/_alignment.py
```diff
@@ -155,6 +171,14 @@ def solve(
             **kwargs,
         )  # type: ignore[return-value]
 
+    def _normalize_spatial(
```
michalk8, line 174: I think this can be defined in a new mixin class to avoid repeating the code (or alternatively in a utility function; have no strong opinion).
In the former case, how about using:
```python
def _normalize_spatial(self, normalize_key: Optional[str] = None) -> str:
   if normalize_key is None:
       return self.spatial_key
   spatial_key = f"{spatial_key}_{normalize_key}"
   self.adata.obsm[spatial_key] = ...
   return spatial_key
```
and later use it as:
```python
self.spatial_key = spatial_key
self.spatial_key = self._normalize_spatial(normalize_key)
```

src/moscot/problems/space/_alignment.py
```diff
@@ -59,6 +62,11 @@ def prepare(
         reference
             Only used if `policy="star"`, it's the value for reference stored
             in :attr:`anndata.AnnData.obs` ``["batch_key"]``.
+        normalize_spatial
```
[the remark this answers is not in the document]
michalk8: Or alternatively, keep the boolean flag always store the standardized data in `adata.obsm[f"{self.spatial_key}_norm"]`.

src/moscot/problems/space/_mapping.py
```diff
@@ -122,6 +130,13 @@ def prepare(
         --------
         %(ex_prepare)s
         """
+        if normalize_spatial and isinstance(spatial_key, str):
+            self._normalize_spatial(
+                spatial_key=spatial_key,
+                normalize_key=normalize_key,
+            )
+            spatial_key = f"{spatial_key}_{normalize_key}"
+            logger.info(f"Normalizing spatial coordinates and saving them in `adata.obsm['{spatial_key}']`.")
```
michalk8, line 139: Would move this logging to the function before any computation happens.

tests/problems/space/test_mixins.py
```diff
@@ -34,8 +34,8 @@ def test_analysis(self, adata_space_rotate: AnnData):
             tgts = set(categories) - set(ref)
             for c in zip(tgts):
                 assert (
-                    adata_ref[adata_ref.obs.batch == c].obsm["spatial_warp"].shape
-                    == adata_ref[adata_ref.obs.batch == c].obsm["spatial"].shape
+                    adata_ref[adata_ref.obs.batch == c].obsm["spatial"].shape
+                    == adata_ref[adata_ref.obs.batch == c].obsm["spatial_norm_warp"].shape
```
michalk8, line 38: Think this should be `spatial_warp_norm` (also in other places). Would also test that the normalized coordinates have 0 mean and 1 std.

michalk8 (changes requested):

src/moscot/problems/space/_alignment.py
MUCDK: btw , in the `TranslationProblem` we expect the user to normalize the data beforehand. Can't we expect the user to do so?

src/moscot/problems/space/_alignment.py
```diff
@@ -74,6 +82,14 @@ def prepare(
         --------
         %(ex_prepare)s
         """
+        if normalize_spatial:
```
MUCDK, line 85: please add a test for this

giovp:
ok I realized that it's more complicated that I thought, as we allow to have batches in target for mapping and ofc for sequential in alignment. Therefore, the normalization should be handled as a callback, but in current implementation, there cannot be more than one run in sequence, e.g. see:

https://github.com/theislab/moscot/blob/8a2bb975f54d71e13ff4319a69051131b6184a40/src/moscot/base/problems/compound_problem.py#L120-L121

would you be ok if I allow for callback to change from 
```python
callback: Optional[Union[Literal["local-pca"], Callback_t]] = None,
```
to
```python
callback: Optional[Union[list[str], Literal["local-pca"], Callback_t]] = None,
```
?

giovp:
ok let's try again @michalk8 @MUCDK 🙏

tests/problems/space/test_mapping_problem.py
```diff
@@ -40,7 +47,11 @@ def test_prepare(self, adata_mapping: AnnData, sc_attr: Mapping[str, str], joint
         assert mp.problems == {}
         assert mp.solutions == {}
 
-        mp = mp.prepare(batch_key="batch", sc_attr=sc_attr, joint_attr=joint_attr)
+        mp = mp.prepare(batch_key="batch", sc_attr=sc_attr, joint_attr=joint_attr, normalize_spatial=normalize_spatial)
+        if normalize_spatial:
+            assert mp[("1", "ref")].x.data_src.std() == mp[("2", "ref")].x.data_src.std() == 1
```
michalk8, line 52: Same as above, would just `np.testing.assert_allclose`.

michalk8 (changes requested):

src/moscot/base/problems/problem.py
```diff
@@ -506,6 +506,25 @@ def concat(x: ArrayLike, y: ArrayLike) -> ArrayLike:
             return {term: TaggedArray(x, tag=Tag.POINT_CLOUD)}
         raise ValueError(f"Expected `term` to be one of `x`, `y`, or `xy`, found `{term!r}`.")
 
+    @staticmethod
+    def _spatial_norm_callback(
+        term: Literal["x", "y"],
```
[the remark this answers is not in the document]
giovp: answering here related to this, the problem is that this is not true, as the spatial dimension will always only gonna be in y or x and not xy, and it doesn't make sense to normalize like this unless with spatial dimensions (e.g. with gexp or pca you'd just standard scale). So I would narrow the signature and maybe raise error if `xy`, wdyt?

src/moscot/base/problems/problem.py
```diff
@@ -506,6 +506,25 @@ def concat(x: ArrayLike, y: ArrayLike) -> ArrayLike:
...
+    ) -> Dict[Literal["x", "y"], TaggedArray]:
+        spatial_key = kwargs["spatial_key"]
+        if term == "x":
+            spatial = adata.obsm[spatial_key]
+        if term == "y":
+            if isinstance(adata_y, AnnData):
+                spatial = adata_y.obsm[spatial_key]
+            else:
+                raise ValueError("When `term` is `y` `adata_y` cannot be `None`.")
+        logger.info(f"Normalizing spatial coordinates of `{term}`.")
```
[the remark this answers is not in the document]
giovp: mmh we have it for pca, would keep it as it's an important log, the user needs to know about this internal normalization

src/moscot/base/problems/problem.py
```diff
@@ -506,6 +506,25 @@ def concat(x: ArrayLike, y: ArrayLike) -> ArrayLike:
...
 
+    @staticmethod
+    def _spatial_norm_callback(
+        term: Literal["x", "y"],
+        adata: AnnData,
+        adata_y: Optional[AnnData] = None,
+        **kwargs: Any,
+    ) -> Dict[Literal["x", "y"], TaggedArray]:
+        spatial_key = kwargs["spatial_key"]
+        if term == "x":
```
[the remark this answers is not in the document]
giovp: see above my reply

giovp:
address all comments etc`xy` as explained in comment above.

src/moscot/base/problems/compound_problem.py
```diff
@@ -118,6 +118,8 @@ def verify_data(data: Mapping[Literal["xy", "x", "y"], TaggedArray]) -> None:
             return {}
         if callback == "local-pca":
             callback = problem._local_pca_callback
+        if callback == "spatial-norm":
```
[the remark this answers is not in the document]
giovp: this doesnt work, see last comment in main conv

giovp:
@michalk8 tests were failing due to 

<details>
<summary>Details</summary>

```
def _callback_handler(
        self,
        term: Literal["x", "y", "xy"],
        key_1: K,
        key_2: K,
        problem: B,
        *,
        callback: Optional[Union[Literal["local-pca"], Callback_t]] = None,
        **kwargs: Any,
    ) -> Mapping[Literal["xy", "x", "y"], TaggedArray]:
        def verify_data(data: Mapping[Literal["xy", "x", "y"], TaggedArray]) -> None:
            keys = ("xy", "x", "y")
            for key, val in data.items():
                if key not in keys:
                    raise ValueError(f"Expected key to be one of `{keys}`, found `{key!r}`.")
                if not isinstance(val, TaggedArray):
                    raise TypeError(f"Expected value for `{key}` to be a `TaggedArray`, found `{type(val)}`.")
    
        if callback is None:
            return {}
        if callback == "local-pca":
            callback = problem._local_pca_callback
        elif callback == "spatial-norm":
            callback = problem._spatial_norm_callback
        else:
>           raise ValueError(f"Unknown callback `{callback!r}`.")
E           ValueError: Unknown callback `<MagicMock name='xy_callback' id='10841556016'>`.

src/moscot/base/problems/compound_problem.py:124: ValueError
=========================================================== short test summary info ===========================================================
FAILED tests/problems/base/test_compound_problem.py::TestCompoundProblem::test_custom_callback_lin - ValueError: Unknown callback `<MagicMock name='xy_callback' id='10841556016'>`.
```

</details>

reverted back to not raise value error

michalk8 (approved):

michalk8:
Thanks a lot @giovp, lgtm!

MUCDK (approved):
LGTM, thanks

Issue #544, opened 2023-05-25
Test for the correct use of parameters of ott-jax which are not exposed in moscot


MUCDK:
We do not expose all arguments in the methods, especially for `solve` methods. Still, the user might want to pass an argument which can be handled by ott-jax. 

We need to test that these are used correctly/can be passed via moscot.

Example: Test that the anderson acceleration, which we do not expose, is passed correctly:

```python
adata = simulate_data(key="day")

tp = TemporalProblem(adata)
tp = tp.prepare("day")
tp = tp.solve(anderson=acceleration.AndersonAcceleration(memory=5, refresh_every=3))

assert isinstance(tp[0,1].solver.solver.anderson, AndersonAcceleration)
```

MUCDK:
please add tests also for 
- another parameter for the linear case (`SinkhornProblem`)
- analogously, choose two parameters for the quadratic case, i.e. for the `GWProblem`, (parameters which are present in OTT, but not in moscot exposed)

PR #554, opened 2023-06-19
testing unexposed ott parameters in generic solvers

ArinaDanilina:
added test_passing_ott_kwargs function in generic sinkhorn tests, part of #544

### DIFF
--- a/tests/problems/generic/test_fgw_problem.py
+++ b/tests/problems/generic/test_fgw_problem.py
@@ -269,6 +269,7 @@ class TestFGWProblem:
         problem = problem.solve(
             max_iterations=1,
             linear_solver_kwargs={
+                "max_iterations": 1,
                 "anderson": acceleration.AndersonAcceleration(memory=memory, refresh_every=refresh),
                 "recenter_potentials": recenter,
             },
--- a/tests/problems/generic/test_fgw_problem.py
+++ b/tests/problems/generic/test_fgw_problem.py
@@ -284,8 +284,8 @@ class TestFGWProblem:
         assert recenter_potentials == recenter
 
     @pytest.mark.parametrize("warm_start", [True, False])
-    @pytest.mark.parametrize("unscale", [True, False])
-    def test_passing_ott_kwargs_quadratic(self, adata_space_rotate: AnnData, warm_start: bool, unscale: bool):
+    @pytest.mark.parametrize("inner_errors", [True, False])
+    def test_passing_ott_kwargs_quadratic(self, adata_space_rotate: AnnData, warm_start: bool, inner_errors: bool):
         problem = GWProblem(adata=adata_space_rotate)
         problem = problem.prepare(
             key="batch",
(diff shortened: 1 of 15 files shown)


ArinaDanilina:
added test_passing_ott_kwargs_linear in generic gw and fgw tests

ArinaDanilina:
added test_passing_ott_kwargs_quadratic in fgw and gw tests, finished #544

tests/problems/generic/test_fgw_problem.py
```diff
@@ -252,3 +253,51 @@ def test_prepare_different_costs(self, adata_time: AnnData):
...
+        problem = GWProblem(adata=adata_space_rotate)
+        problem = problem.prepare(
+            key="batch",
+            policy="sequential",
+            joint_attr="X_pca",
+            x_attr={"attr": "obsm", "key": "spatial"},
+            y_attr={"attr": "obsm", "key": "spatial"},
+        )
+
+        problem = problem.solve(
```
MUCDK, line 269: How long does this take locally? 
As we can test this without convergence of the solver, we can set `max_iterations=1`.

MUCDK (changes requested):
see comments

tests/problems/generic/test_fgw_problem.py
ArinaDanilina: with max_iterations=1 it takes 1 min

ArinaDanilina: in Sinkhorn it takes 20 sec

ArinaDanilina:
max_iterations=1 speeds it up 1,5 times, in Sinkhorn it now takes 20 sec and the longest test in fgw is 1 min

tests/problems/generic/test_fgw_problem.py
MUCDK: ok can you also set the number of inner iterations to a smaller value please? E.g. also 1

MUCDK (changes requested):
see comment

tests/problems/generic/test_fgw_problem.py
ArinaDanilina: done, down to 40 sec

tests/problems/generic/test_sinkhorn_problem.py
```diff
@@ -166,3 +167,26 @@ def test_pass_arguments(self, adata_time: AnnData, args_to_check: Mapping[str, A
...
+    @pytest.mark.parametrize(("memory", "refresh"), [(1, 1), (5, 3), (7, 5)])
+    @pytest.mark.parametrize("recenter", [True, False])
+    def test_passing_ott_kwargs(self, adata_time: AnnData, memory: int, refresh: int, recenter: bool):
+        problem = SinkhornProblem(adata=adata_time)
+        problem = problem.prepare(
+            key="time",
+            policy="sequential",
+        )
+
+        problem = problem.solve(
```
MUCDK, line 180: inner iterations = 1

MUCDK (changes requested):
see comments

tests/problems/generic/test_sinkhorn_problem.py
MUCDK: are you sure this is how to pass inner iterations? I think it's via `linear_solver_kwargs`, but not entirely sure. Please double-check.

MUCDK (changes requested):
see comment

tests/problems/generic/test_sinkhorn_problem.py
ArinaDanilina: here it's a SinkhornProblem so inner iterations get passed just like that. In GWProblem I do pass it in ```linear_solver_kwargs={
                "inner_iterations": 1,
                "max_iterations": 1,
            }```

MUCDK (approved):
LGTM, thanks!

Btw, `inner_iterations` determines how often we log the error in Sinkhorn. So setting this here doesn't have an effect / is unnecessary, but obv. fine.

PR #557, opened 2023-06-21
Deps/update ott

MUCDK:


giovp:
I think this can be closed for now?

src/moscot/backends/ott/solver.py
```diff
@@ -141,7 +141,7 @@ def __init__(
         self,
         jit: bool = True,
         rank: int = -1,
-        epsilon: float = 0.0,
+        epsilon: Optional[float] = 0.0,
```
michalk8, line 144: This cannot be optional for LR solvers.

tests/problems/space/test_mixins.py
```diff
@@ -51,6 +51,7 @@ def test_analysis(self, adata_space_rotate: AnnData):
                     == adata_ref[adata_ref.obs.batch == c].obsm["spatial"].shape
                 )
 
+    @pytest.mark.skip("TODO(giovp): regenerate solutions")
```
michalk8, line 54: These should be regenerated now.

michalk8 (changes requested):
Looks good, just minor comments.

Issue #562, opened 2023-06-22
add usage example in subset_policy.py


MUCDK:
@ArinaDanilina please open issue and self-assign

_Originally posted by @MUCDK in https://github.com/theislab/moscot/pull/515#discussion_r1238609827_

Issue #561, opened 2023-06-22
link policy example in subset_policy


MUCDK:
@ArinaDanilina new issue for this please

_Originally posted by @MUCDK in https://github.com/theislab/moscot/pull/515#discussion_r1238612050_

PR #572, opened 2023-06-28
Added links to example in subset policy

ArinaDanilina:
#561 and #562


ArinaDanilina:
newer PR was created with branch from main (pr #582)

Issue #604, opened 2023-09-13
`cell_transition` changed


MUCDK:
At the moment, we cannot reproduce the multimodality tutorial, as the `cell_transition` doesn't work any more. 

A non-scalable work around is the following code

```
def aggregate_transport_matrix(adata_source, adata_target, tmat, aggregation_key = "celltype", forward = True):
    df_source = adata_source.obs[aggregation_key]
    df_target = adata_target.obs[aggregation_key]

    annotations_source = adata_source.obs[aggregation_key].cat.categories
    annotations_target = adata_target.obs[aggregation_key].cat.categories

    tm = pd.DataFrame(
        np.zeros((len(annotations_source), len(annotations_target))),
        index=annotations_source,
        columns=annotations_target,
    )
    
    for annotation_src in annotations_source:
        for annotation_tgt in annotations_target:
            tm.loc[annotation_src, annotation_tgt] = tmat[
                np.ix_((df_source == annotation_src).squeeze(), (df_target == annotation_tgt).squeeze())
            ].sum()
    return tm.div(tm.sum(axis=int(forward)), axis=int(not forward))
```

tm = aggregate_transport_matrix(adata_atac, adata_rna, ftp["src", "tgt"].solution.transport_matrix, forward=True, aggregation_key="cell_type")

PR #605, opened 2023-09-24
fix `cell_transition`

ArinaDanilina:
the multimodality tutorial works as expected #604


MUCDK:
@giovp do you understand why this fixes the bug? did the dispatching behaviour change?

giovp:
@MUCDK I have no clue tbh ...

PR #626, opened 2023-11-07
Fix pandas 2 error with lineage tree loading

### LINKED ISSUE #625: Dataset loading for moslin fails
Due to a pandas 2 problem, calling `adata, tree = moscot.datasets.c_elegans()` gives 

```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[18], line 1
----> 1 adata, full_reference_tree = mc.datasets.c_elegans(DATA_DIR / "c_elegans")

File ~/Projects/moscot/src/moscot/datasets.py:183, in c_elegans(path, force_download, **kwargs)
    181 # TODO(michalk8): also cache or store in AnnData ad Newick + reconstruct?
    182 with urllib.request.urlopen("https://figshare.com/ndownloader/files/39943603") as fin:
--> 183     tree = pickle.load(fin)
    185 return adata, tree

ModuleNotFoundError: No module named 'pandas.core.indexes.numeric'
```

It would indeed by better to write the tree as a Newick to AnnData and reconstruct it from there, that would probably be more stable @michalk8

Marius1311:
Hi @michalk8, this is one way of fixing an error that occurred to me when trying to read in the C. elegans lineage trees via `adata, tree = moscot.datasets.c_elegans()`, see https://github.com/theislab/moscot/issues/625. 

The problem was that node information was stored in DataFrames, which somehow caused a problem in pandas 2. I now convert dataframes to dict and save the tree in `gml` format. 

Ideally, the tree would also be cached, just like the AnnData object, but this might be fine for now?

closes #625

### DIFF
--- a/src/moscot/datasets.py
+++ b/src/moscot/datasets.py
@@ -178,9 +178,9 @@ def c_elegans(
         force_download=force_download,
         **kwargs,
     )
-    # TODO(michalk8): also cache or store in AnnData ad Newick + reconstruct?
-    with urllib.request.urlopen("https://figshare.com/ndownloader/files/39943603") as fin:
-        tree = pickle.load(fin)
+
+    with urllib.request.urlopen("https://figshare.com/ndownloader/files/43064842") as fin:
+        tree = nx.read_gml(fin)
 
     return adata, tree
 


giovp:
@Marius1311 if the test pass, we can merge!

giovp (approved):

PR #627, opened 2023-11-07
add cell-wise entropy computation

MUCDK:


MUCDK:
This should be merged after #633

PR #633, opened 2023-12-01
start geodesic implementation with tests

MUCDK:


MUCDK:
TODO: add tests for quadratic OT & multiple OT problems per CompoundProblem & adapt docstrings.

MUCDK:
need to fix `GraphOTTOutput._apply`, as it currently always returns zeros.

src/moscot/backends/ott/solver.py
```diff
@@ -192,7 +214,10 @@ def _prepare(
         )
         if cost_matrix_rank is not None:
             geom = geom.to_LRCGeometry(rank=cost_matrix_rank)
-        self._problem = linear_problem.LinearProblem(geom, **kwargs)
+        if xy.is_graph:
+            a = jnp.concatenate((a, jnp.zeros_like(self._b)), axis=0)
+            b = jnp.concatenate((jnp.zeros_like(self._a), b), axis=0)
```
giovp, line 219: `self._a` and `self._b` are used only here or also somewhere downstream? Because If it's only here, why not getting the shape of the left and right marginal from the geometry `geom` ?

src/moscot/backends/ott/solver.py
```diff
@@ -284,24 +311,27 @@ def _prepare(
         alpha: float = 0.5,
         **kwargs: Any,
     ) -> quadratic_problem.QuadraticProblem:
+        self._a = a
+        self._b = b
```
giovp, line 315: also here, not sure I understand why they are needed

src/moscot/backends/ott/solver.py
MUCDK: we cannot get the shape from the `geom` because the `geom` is of shape `[n+m, n+m]`

MUCDK:
> minor things, approving in advance. One question, could you explain again why we might (or might not) need the set_x_graph/set_y_graph ? you mentioned it in some meeting but forgot.

A graph in `x` or `y` is of shape `[n,n]` or `[m,m]`, hence we can pass a pd.DataFrame of that shape, while in `xy` the shape is `[n+m, n+m]`.

An argument for a separate `set_x_graph` would be the data type which can be passed, if we would like to be able to pass the graph as a `csr_matrix`, and hence not as `pd.DataFrame` as in `set_x`. But we have to instantiate the cost matrix anyways, hence there might be no point in allowing to pass as something else than a `pd.DataFrame`

src/moscot/base/problems/problem.py
```diff
@@ -249,19 +250,22 @@ def __init__(
         self._a: Optional[ArrayLike] = None
         self._b: Optional[ArrayLike] = None
 
+        self._t: Optional[float] = None  # only needed for diffusion-based distances
```
michalk8, line 253: This feels like a design mistake, since it can be set via `set_graph_xy`. Is there an easy way to refactor this?

src/moscot/backends/ott/solver.py
```diff
@@ -43,14 +44,19 @@ def __init__(self, jit: bool = True):
         self._solver: Optional[OTTSolver_t] = None
         self._problem: Optional[OTTProblem_t] = None
         self._jit = jit
+        self._a: Optional[jnp.ndarray] = None
+        self._b: Optional[jnp.ndarray] = None
+        self._requires_graph_output = False  # resolve which output to use, OTTOutput v. GraphOTTOutput
```
michalk8, line 49: Remove this please, not necessary - can be done via `isinstance` check later, would do
```def _solver_output_type(self, prob: ....) -> Union[Type[...], Type[...]]:
   # if a geodesic, require graph output, otherwise normal.
```

src/moscot/backends/ott/solver.py
```diff
@@ -114,6 +155,11 @@ def is_low_rank(self) -> bool:
         """Whether the :attr:`solver` is low-rank."""
         return self.rank > -1
 
+    @property
+    def graph_in_linear_term(self) -> bool:
```
michalk8, line 159: Where is this used? Imho the solver doesn't need to care about this and can be removed.

src/moscot/backends/ott/solver.py
```diff
@@ -175,24 +223,31 @@ def _prepare(
...
             batch_size=batch_size,
+            n_src_tgt=(len(self._a), len(self._b)),
             scale_cost=scale_cost,
+            t=t,
             **cost_kwargs,
         )
         if cost_matrix_rank is not None:
             geom = geom.to_LRCGeometry(rank=cost_matrix_rank)
-        self._problem = linear_problem.LinearProblem(geom, **kwargs)
+        if self._requires_graph_output:
```
michalk8, line 247: Use instance check on geom or similar.

src/moscot/backends/ott/solver.py
```diff
@@ -88,15 +94,50 @@ def _create_geometry(
             return geometry.Geometry(
                 kernel_matrix=arr, epsilon=epsilon, relative_epsilon=relative_epsilon, scale_cost=scale_cost
             )
+        if x.is_graph:  # we currently only support this for the linear term.
+            if x.cost == "geodesic":
+                if self.problem_kind == "linear":
+                    if t is None:
+                        self._requires_graph_output = True
```
michalk8, line 101: Remove this please.

michalk8 (changes requested):

src/moscot/problems/generic/_generic.py
```diff
@@ -41,7 +49,7 @@ def prepare(
         key: str,
         joint_attr: Optional[Union[str, Mapping[str, Any]]] = None,
         policy: Literal["sequential", "explicit", "star"] = "sequential",
-        cost: OttCostFn_t = "sq_euclidean",
+        cost: Union[OttCostFn_t, Literal["geodesic"]] = "sq_euclidean",
```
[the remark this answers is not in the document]
MUCDK: At the moment we don't support it for quadratic terms, hence we have to distinguish between those. Maybe it's then a good idea to allow for geodesic distances also in the quadratic terms now. @giovp wdyt?

src/moscot/base/problems/problem.py
```diff
@@ -573,39 +577,61 @@ def concat(x: ArrayLike, y: ArrayLike) -> ArrayLike:
...
+                raise ValueError("When `term` is `xy`, `adata_y` cannot be `None`.")
+            if use_rep not in adata.obsm:
+                raise ValueError(f"Unable to find `{use_rep}` in `adata.obsm`.")
+            if use_rep not in adata_y.obsm:
+                raise ValueError(f"Unable to find `{use_rep}` in `adata_y.obsm`.")
+            adata_concat = ad.concat((adata, adata_y), join="inner")
+            logger.info(f"Computing graph construction for `xy` using `{use_rep}`")
+            sc.pp.neighbors(adata_concat, **kwargs)
+            return TaggedArray(
+                data_src=adata_concat.obsp["connectivities"].astype("float64"), data_tgt=None, tag=Tag.GRAPH
```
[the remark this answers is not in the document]
MUCDK: Simply put, this is the only way it works. 

Connectivities are always stored as `f32`. 
Might be smarter to infer it, how would you do it?

src/moscot/base/problems/problem.py
MUCDK: I discussed this with @giovp, as I didn't find an easy solution.

One option would be to instantiate the cost matrix in `set_graph_xy` if `t is not None` (because then `t!=eps/4` and we need to instantiate the cost matrix). The limitation would be that in (F)GW problems, we need to instantiate the cost matrix anyways, which currently doesn't happen in `set_graph_xy` because we can also directly use the geodesic cost, e.g. via the `graph-construction` callback in `prepare`. 

I would also be fine with instantiating it in two different places, but @giovp you were against it, right

src/moscot/backends/ott/solver.py
```diff
@@ -43,14 +44,19 @@ def __init__(self, jit: bool = True):
...
         self._problem: Optional[OTTProblem_t] = None
         self._jit = jit
+        self._a: Optional[jnp.ndarray] = None
+        self._b: Optional[jnp.ndarray] = None
+        self._requires_graph_output = False  # resolve which output to use, OTTOutput v. GraphOTTOutput
 
     def _create_geometry(
         self,
         x: TaggedArray,
+        n_src_tgt: Tuple[int, int],
```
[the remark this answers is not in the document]
MUCDK: fair. would prefer to call it `problem_shape`, because it's explicitly not the graph shape, but the problem shape (graph_shape > problem_shape)

src/moscot/base/problems/problem.py
```diff
@@ -657,6 +683,54 @@ def estimate_marginals(self, adata: AnnData, *, source: bool, **kwargs: Any) ->
...
+            pd.testing.assert_series_equal(expected_series, data.columns.to_series())
+            data_src = data.to_numpy()
+        elif isinstance(data, tuple):
+            pd.testing.assert_series_equal(expected_series, data[1])
+            pd.testing.assert_series_equal(expected_series, data[2])
+            data_src = data[0]
+        else:
+            raise ValueError(
+                "Expected data to be a pd.DataFrame or a tuple of (sp.csr_matrix, pd.Series, pd.Series), "
+                + f"found {type(data)}."
```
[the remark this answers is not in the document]
MUCDK: wdym by "+ use `type(data)"?

src/moscot/backends/ott/solver.py
```diff
@@ -88,17 +93,62 @@ def _create_geometry(
...
+                    return geodesic.Geodesic.from_graph(
+                        arr, t=epsilon / 4.0, directed=kwargs.pop("directed", True), **kwargs
+                    )
+
+                n_src, n_tgt = problem_shape
+                if n_src + n_tgt != arr.shape[0]:
+                    raise ValueError(f"Expected `x` to have `{n_src + n_tgt}` points, found `{arr.shape[0]}`.")
+                cm = geodesic.Geodesic.from_graph(
+                    arr, t=t, directed=kwargs.pop("directed", True), **kwargs
+                ).cost_matrix[:n_src, n_src:]
```
giovp, line 136: why do you instantiate cost? it's not in the quadratic case right?

giovp: actually the whole bit above. I don't see why it's needed? the quadratic case is below

src/moscot/problems/time/_lineage.py
```diff
@@ -44,7 +44,7 @@ def prepare(
         time_key: str,
         joint_attr: Optional[Union[str, Mapping[str, Any]]] = None,
         policy: Literal["sequential", "triu", "tril", "explicit"] = "sequential",
-        cost: OttCostFn_t = "sq_euclidean",
+        cost: Union[OttCostFn_t, Literal["geodesic"]] = "sq_euclidean",
```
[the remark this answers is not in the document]
MUCDK: see comment above. atm not supported for the quadratic term, hence we cannot add it to `OttCostFn_t` , as we also use it for quadratic cost functions.

giovp (changes requested):
couple more things that I don't get from `create_graph_geometry`

src/moscot/backends/ott/solver.py
```diff
@@ -165,6 +211,8 @@ def __init__(
 
     def _prepare(
         self,
+        a: jnp.ndarray,
```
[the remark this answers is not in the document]
MUCDK: This is only possible if we don't have a graph geometry. Hence, we would have to distinguish the cases. Below, we would have to do: 

```
if isinstance(geom, geodesic.Geodesic):
     assert a is not None
     assert b is not None
     a = jnp.concatenate((a, jnp.zeros_like(self._b)), axis=0)
     b = jnp.concatenate((jnp.zeros_like(self._a), b), axis=0)
```
We cannot infer the shape of the problem at this stage. Hence, the shape must have been inferred before, and the marginals created accordingly.


What do you think?

src/moscot/base/problems/problem.py
giovp: so the problem is I think that there should only one place where the geodesic is instantitate, as a cost or as a geometry, and it should be in solve. Now I understand that the problem is that that step can be called by either the prepare (and hence instantiate the OTPRoblem) or the `set_graph_xy`. I am not sure what would be an easy solution for this, maybe we can think about behaviour rather than API as a guiding principle?

For example, my understanding is that, if the user wants to use the geodesic cost, it has:
- pass it to prepare and set t, in case of quadratic problem
- use set_graph_xy explicitly, in case of linear problem

Is that correct>?

MUCDK: - in case of the quadratic problem, the user should still be able to use `set_graph_xy`
- in case of the linear problem, the user can also pass it via the `prepare` method.

The main difference is that
- if the user goes via the `prepare` method, we automatically set `4t=eps`
- if the user goes via the `set_graph_xy` method, `t` and `eps` are handled independently of each other

src/moscot/backends/ott/solver.py
MUCDK: If `t!=eps/4` we have to instantiate it.

src/moscot/utils/tagged_array.py
```diff
@@ -82,7 +98,7 @@ def from_adata(
         attr: Literal["X", "obsp", "obsm", "layers", "uns"],
         tag: Tag = Tag.POINT_CLOUD,
         key: Optional[str] = None,
-        cost: CostFn_t = "sq_euclidean",
+        cost: Union[CostFn_t, Literal["geodesic"]] = "sq_euclidean",
```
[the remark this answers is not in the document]
MUCDK: see comments above, currently `geodesic` cannot be added to `OTTCostFn_t`

tests/backends/ott/test_backend.py
```diff
@@ -39,7 +39,7 @@ def test_matches_ott(self, x: Geom_t, eps: Optional[float], jit: bool):
         assert solver.xy is None
         assert isinstance(solver.solver, Sinkhorn)
 
-        pred = solver(xy=(x, x), epsilon=eps)
+        pred = solver(a=jnp.ones(len(x)) / len(x), b=jnp.ones(len(x)) / len(x), xy=(x, x), epsilon=eps)
```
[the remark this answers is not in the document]
MUCDK: see comment above about where to infer marginals.

src/moscot/base/problems/problem.py
```diff
@@ -337,30 +341,29 @@ def prepare(
...
-            self._xy = xy if isinstance(xy, TaggedArray) else self._handle_linear(**xy)
-        elif x and y and not xy:
-            self._problem_kind = "quadratic"
-            if isinstance(x, TaggedArray):
-                self._x = x
+        if xy:
+            if "tagged_array" in xy:
+                kws, _= self._split_xy_kwargs(**xy)
+                xy=dict(xy)
+                self._xy = xy.pop("tagged_array")._add_cost(**kws)
```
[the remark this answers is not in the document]
MUCDK: It's true for the `xy` term, but we cannot do it for the `x` and `y` as we don't create a new kwargs variable as we do here from `xy`.

src/moscot/problems/time/_lineage.py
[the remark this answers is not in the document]
MUCDK: now resolved with allowing in quad. term.

src/moscot/utils/tagged_array.py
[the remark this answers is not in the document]
MUCDK: now resolved with allowing in quad. term.

src/moscot/problems/generic/_generic.py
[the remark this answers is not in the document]
MUCDK: now resolved with allowing in quad. term.

giovp (approved):

src/moscot/base/problems/problem.py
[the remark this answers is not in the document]
MUCDK: add TODO

src/moscot/backends/ott/solver.py
```diff
@@ -175,24 +253,32 @@ def _prepare(
         scale_cost: Scale_t = 1.0,
         cost_kwargs: Mapping[str, Any] = types.MappingProxyType({}),
         cost_matrix_rank: Optional[int] = None,
+        ts: Tuple[Optional[float], Optional[float], Optional[float]] = (None, None, None),
```
MUCDK, line 256: make NameTuple

Issue #632, opened 2023-12-01
Write tests for `moscot.costs`


MUCDK:
From codecov we can see that `moscot.costs` has a very low test coverage (https://app.codecov.io/gh/theislab/moscot/tree/main/src%2Fmoscot%2Fcosts). Hence, we need a few tests here.

PR #637, opened 2023-12-11
Cost Tests

selmanozleyen:
Hi!

This PR is meant to solve #632,

Currently I am testing the initialization of the objects, the special cases, and functionality with only using sample input and expected outputs.

Remaining tasks besides the future feedback:
- test initialization of LeafDistance
- test LeafDistance with sample input and output

ping: @giovp , @MUCDK

### DIFF
--- a/src/moscot/costs/_utils.py
+++ b/src/moscot/costs/_utils.py
@@ -30,10 +30,7 @@ def get_available_costs(backend: Optional[str] = None) -> Dict[str, Tuple[str, .
     -------
     Dictionary with keys as backend names and values as registered cost functions.
     """
-    groups: Dict[str, List[str]] = collections.defaultdict(list)
-    for key in _REGISTRY:
-        back, *name = key.split(_SEP)
-        groups[back].append(_SEP.join(name))
+    groups: Dict[str, List[str]] = _get_available_backends_n_costs()
 
     if backend is None:
         return {k: tuple(v) for k, v in groups.items()}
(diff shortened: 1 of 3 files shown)


tests/costs/test_cost.py
```diff
@@ -1,2 +1,76 @@
...
+        # Check if the computed distance matches the expected distance
+        np.testing.assert_almost_equal(computed_distance, expected_distance, decimal=4)
+
+    @staticmethod
+    def test_scaled_hamming_dist_if_nan():
+        # Sample input arrays with no shared indices
+        x = np.array([-1, -1, 0, 1])
+        y = np.array([0, 1, -1, -1])
+
+        assert _scaled_hamming_dist(x, y) is np.nan
```
MUCDK, line 53: Can we adapt the function such that it raises a `ValueError`? Otherwise we'll have `NaN` values in the cost matrix, which will lead to errors which are harder to trace back.

MUCDK (changes requested):
Thanks @selmanozleyen . see minor comments

src/moscot/costs/_utils.py
```diff
@@ -46,3 +43,17 @@ def get_available_costs(backend: Optional[str] = None) -> Dict[str, Tuple[str, .
 def register_cost(name: str, *, backend: str) -> Any:
     """Register cost function for a specific backend."""
     return _REGISTRY.register(f"{backend}{_SEP}{name}")
+
+
+def _get_available_backends_n_costs():
```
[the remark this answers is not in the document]
selmanozleyen: you are right, just an old hobbit

tests/costs/test_cost.py
selmanozleyen: done

selmanozleyen:
Ok, I also added tests for leaf_distance which should cover everything except one case in `_get_leaves` method. I am not sure if I should add a test for it because it is a private method and the argument it takes is never used.

MUCDK (approved):

PR #640, opened 2023-12-15
fix spatial callback

giovp:


giovp:
@MUCDK all the mapping tests fail here:

```
name = 'sq_euclidean', backend = 'ott', kwargs = {'attr': 'obsm', 'key': 'spatial'}, key = 'ott-sq_euclidean'

    def get_cost(name: str, *, backend: str = "moscot", **kwargs: Any) -> Any:
        """Get cost function for a specific backend."""
        key = f"{backend}{_SEP}{name}"
        if key not in _REGISTRY:
            raise ValueError(f"Cost `{name!r}` is not available for backend `{backend!r}`.")
>       return _REGISTRY[key](**kwargs)
E       TypeError: SqEuclidean() takes no arguments
```

that is because of the introduction of this
https://github.com/theislab/moscot/blob/bb9be0407082cde876235dad3356546fb7e9bc1e/src/moscot/base/problems/problem.py#L342-L354

PR #642, opened 2023-12-18
spatial geodesic

giovp:


src/moscot/backends/ott/solver.py
```diff
@@ -97,6 +98,11 @@ def _create_geometry(
             if is_linear_term:
                 self._graph_in_linear_term = True
             if x.cost == "geodesic":
+                if n_src_tgt is not None:
+                    n_src, n_tgt = n_src_tgt
+                    if n_src + n_tgt != arr.shape[0]:
+                        raise ValueError(f"Expected `x` to have `{n_src + n_tgt}` points, found `{arr.shape[0]}`.")
+                    arr = arr[:n_src, n_src:]
```
giovp, line 105: @MUCDK do you think this makes sense? I think yes cause the indices are checked here https://github.com/theislab/moscot/blob/90600424297d6ec612756d9d2f52c95463c90ea4/src/moscot/base/problems/problem.py#L708

MUCDK: I don't think it makes sense because of another reason: https://github.com/ott-jax/ott/issues/479#issuecomment-1860887301

tests/problems/space/test_alignment_problem.py
```diff
@@ -122,49 +123,51 @@ def test_solve_unbalanced(self, adata_space_rotate: AnnData):
...
+                    index=adata_subset.obs_names,
+                    columns=adata_subset.obs_names,
+                    data=adata_subset.obsp["connectivities"].A.astype("float64"),
+                )
+            )
+
+        ap: AlignmentProblem = AlignmentProblem(adata=adata_space_rotate)
+        ap = ap.prepare(batch_key=batch_column, joint_attr={"attr": "obsm", "key": "X_pca"})
+
+        ap[("0", "1")].set_graph_xy(dfs[0], cost="geodesic")
```
MUCDK, line 155: @giovp why does this work now? I thought ott-jax cannot handle geodesic in FGW?

giovp: it's handled here as you suggested https://github.com/theislab/moscot/blob/f38ea663b2887b94878e740072d5ebaad0fbd6d8/src/moscot/backends/ott/solver.py#L105-L107

MUCDK: ok great!

src/moscot/backends/ott/solver.py
```diff
@@ -97,6 +98,14 @@ def _create_geometry(
             if is_linear_term:
                 self._graph_in_linear_term = True
```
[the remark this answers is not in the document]
giovp: > I think we should now check only for graph in linear term AND cost matrix is not instantiated

I think this is a sound option but I am not sure I understand how a geometry can both be graph and be a cost_matrix. since `x.is_cost_matrix` is handled above.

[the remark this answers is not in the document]
MUCDK: The reason why we need the `_graph_in_linear_term` is to indicate whether we need an `OTTOutput` or a `GraphOTTOutput`. If we instantiate the geodesic cost matrix, we need the `OTTOutput`, hence, internally, this is handled as if we had a cost matrix.

PR #647, opened 2024-01-20
passing batch_size fix

ArinaDanilina:


src/moscot/problems/cross_modality/_mixins.py
```diff
@@ -216,6 +218,9 @@ def annotation_mapping(  # type: ignore[misc]
             Key identifying the source distribution.
         target
             Key identifying the target distribution.
+        batch_size
+            Number of rows/columns of the cost matrix to materialize during :meth:`push` or :meth:`pull`.
```
MUCDK, line 222: Please add what happens if batch size is `None`

src/moscot/problems/space/_mixins.py
```diff
@@ -313,6 +315,9 @@ def annotation_mapping(  # type: ignore[misc]
             Key identifying the source distribution.
         target
             Key identifying the target distribution.
+        batch_size
```
MUCDK, line 318: Please add what happens if batch size is `None`

src/moscot/problems/space/_mixins.py
```diff
@@ -648,6 +657,9 @@ def annotation_mapping(  # type: ignore[misc]
...
```
MUCDK, line 660: Please add what happens if batch size is `None`

src/moscot/problems/time/_mixins.py
```diff
@@ -247,7 +247,9 @@ def annotation_mapping(
         forward: bool,
         source: K,
         target: K,
+        batch_size: int | None = None,
```
MUCDK, line 250: here typing is different than in other place

src/moscot/problems/time/_mixins.py
```diff
@@ -269,6 +271,9 @@ def annotation_mapping(
             Key identifying the source distribution.
         target
             Key identifying the target distribution.
+        batch_size
```
MUCDK, line 274: Please add what happens if batch size is `None`

MUCDK (changes requested):
minor changes, then we can merge. thanks

MUCDK (changes requested):
You write everywhere 
```
Number of rows/columns of the cost matrix to 
materialize during :meth:`push` or :meth:`pull`.
If :obj:`None`, the entire cost matrix will be materialized.
Larger value will require more memory.
```

Please swap the 3rd with the 4th line. First talk about non-`None`, and then about `None`.

MUCDK (approved):
Thanks!

PR #649, opened 2024-01-29
mtp.push/pull for all problems

ArinaDanilina:
making plotting push/pull work with all problems


MUCDK (commented):
Thanks Arina!
Is there a way to unify `temporal_key` and `generic_key`?

Also, can we rename `_plot_temporal` to e.g. `_plot_scatter`?

MUCDK (approved):

Issue #532, opened 2023-04-26
Expose the `scaling` parameter in `estimate_marginals`.


MUCDK:

MUCDK:
Referring to documentation.

MUCDK:
@ArinaDanilina then please also make sure to change the example notebook.

PR #653, opened 2024-02-02
expose scaling in estimate_marginals

ArinaDanilina:
#532

### DIFF
--- a/src/moscot/base/problems/birth_death.py
+++ b/src/moscot/base/problems/birth_death.py
@@ -165,6 +165,7 @@ class BirthDeathProblem(BirthDeathMixin, OTProblem):
         source: bool,
         proliferation_key: Optional[str] = None,
         apoptosis_key: Optional[str] = None,
+        scaling: Optional[float] = 1,
         **kwargs: Any,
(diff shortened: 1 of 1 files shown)


src/moscot/base/problems/birth_death.py
```diff
@@ -183,6 +184,8 @@ def estimate_marginals(
             Key in :attr:`~anndata.AnnData.obs` where proliferation scores are stored.
         apoptosis_key
             Key in :attr:`~anndata.AnnData.obs` where apoptosis scores are stored.
+        scaling
```
MUCDK, line 187: Can you please elabore a little bit on what this means? Not as long as in the corresponding example, but the docstrings here are fairly short, so we can spend a little space on this.

MUCDK: I.e. specify what happens if it's `None`, and if it's `float`

MUCDK (changes requested):

MUCDK (changes requested):
Thanks, now please link the example notebook, and then we can merge.

PR #659, opened 2024-02-19
fix `compute_entropy`

giovp:


src/moscot/base/problems/_utils.py
```diff
@@ -721,5 +721,5 @@ def _get_n_cores(n_cores: Optional[int], n_jobs: Optional[int]) -> int:
     return n_cores
 
 
-def _compute_conditional_entropy(p_xy: ArrayLike) -> ArrayLike:
+def _compute_conditional_entropy(p_xy: ArrayLike, c: float = 0.0) -> ArrayLike:
```
MUCDK, line 724: We can remove this function then , right?

src/moscot/base/problems/_mixins.py
```diff
@@ -685,12 +685,18 @@ def compute_entropy(
...
+            Constant added to each row of the transport matrix to avoid numerical instability.
+        kwargs
+            Kwargs for :func:`~scipy.stats.entropy`.
 
         Returns
         -------
         :obj:`None` if ``key_added`` is not None. Otherwise, returns a data frame of shape ``(n_cells, 1)`` containing
         the conditional entropy per cell.
         """
+        from scipy import stats
```
MUCDK, line 698: why not import at the top of the script

src/moscot/base/problems/_utils.py
giovp: yes done

src/moscot/base/problems/_mixins.py
giovp: only used by one test, think it's fine to leave it there?

giovp:
@MUCDK update to revert scipy so that it's supported by 3.8. Yet since scipy 1.12 doesn't support 3.8 anymore, I think we should remove it completely from CI eventually

MUCDK (approved):
LGTM

PR #662, opened 2024-02-26
(feat): Linear GENOT

ilan-gold:
We add two main new classes here:

1. `GENOTLinSolver` wraps the underlying OTT-Jax solver in a moscot-usable way
2. `GENOTLinProblem` at the next level of abstraction up, allowing users to work with `AnnData` objects directly while also using other parts of the moscot API like policies

To run this PR, simply do `pip install -e .` as it points directly at the corresponding OTT-JAX branch.

TODOs:

- [x] Finalize status with OTT-Jax PR
- [x] Remove more dead code as this was based on the ICNN implementation (if there is any)
- [x] Maybe add more tests for certain things that I did not cover?  I will need to look this over
- [x] Maybe a small notebook?
- [x] Figure out how to squash all these commits.  It should be clear where I picked up and the ICNN branch stopped but in any case...would be nice on the eyes.  And we definitely want to squash this PR when merged


pyproject.toml
```diff
@@ -55,15 +55,23 @@ dependencies = [
     "scanpy>=1.9.3",
     "wrapt>=1.13.2",
     "docrep>=0.3.2",
-    "ott-jax>=0.4.5",
+    "ott-jax[neural]@git+https://github.com/ott-jax/ott",
```
ilan-gold, line 58: Oh also we need a release

ilan-gold:
https://github.com/theislab/moscot/actions/runs/9221997646/job/25372233453?pr=662 complains that it can't find `diffrax` with the minimum version from `ott-jax[neural]` on python 3.8, and I'm not sure what to make of that since `ott-jax` tests against 3.8.  That being said, `diffrax` at the minimum version does not support 3.8 it seems: https://github.com/patrick-kidger/diffrax/blob/v0.4.1/setup.py

Thoughts on this from the Kleins? @MUCDK @michalk8

MUCDK:
> https://github.com/theislab/moscot/actions/runs/9221997646/job/25372233453?pr=662 complains that it can't find `diffrax` with the minimum version from `ott-jax[neural]` on python 3.8, and I'm not sure what to make of that since `ott-jax` tests against 3.8. That being said, `diffrax` at the minimum version does not support 3.8 it seems: https://github.com/patrick-kidger/diffrax/blob/v0.4.1/setup.py
> 
> Thoughts on this from the Kleins? @MUCDK @michalk8
> 
> I may be missing something from the testing setup for `ott-jax`, like it doesn't run `neural` against 3.8. But figured it would be quicker to ask.

what about deprecating python 3.8 , also @giovp ?

giovp:
yes lets deprecate!

src/moscot/backends/ott/_utils.py
```diff
@@ -52,6 +62,17 @@ def sinkhorn_divergence(
...
 
+@partial(jax.jit, static_argnames=["k"])
+def get_nearest_neighbors(
+    input_batch: jnp.ndarray, target: jnp.ndarray, k: int = 30
+) -> Tuple[jnp.ndarray, jnp.ndarray]:
+    """Get the k nearest neighbors of the input batch in the target."""
+    if target.shape[0] < k:
+        raise ValueError(f"k is {k}, but must be smaller or equal than {target.shape[0]}.")
+    pairwise_euclidean_distances = pointcloud.PointCloud(input_batch, target).cost_matrix
+    return jax.lax.approx_min_k(pairwise_euclidean_distances, k=k, recall_target=0.95, aggregate_to_topk=True)
```
MUCDK, line 73: can we make `recall_Target` and `aggregate_to_topk` kwargs?

src/moscot/backends/ott/solver.py
```diff
@@ -457,3 +500,199 @@ def _call_kwargs(cls) -> Tuple[Set[str], Set[str]]:
...
+        # TODO(ilan-gold): What are reasonable defaults here?
+        neural_vf = VelocityField(
+            output_dims=[*self._neural_kwargs.get("velocity_field_output_dims", []), target_dim],
+            condition_dims=(
+                self._neural_kwargs.get("velocity_field_condition_dims", [source_dim + condition_dim])
+                if is_conditional
+                else None
+            ),
+            hidden_dims=self._neural_kwargs.get("velocity_field_hidden_dims", [512, 512, 512]),
+            time_dims=self._neural_kwargs.get("velocity_field_time_dims", None),
```
MUCDK, line 612: default: 1024

src/moscot/backends/ott/solver.py
```diff
@@ -457,3 +500,199 @@ def _call_kwargs(cls) -> Tuple[Set[str], Set[str]]:
...
+        condition_dim = self._neural_kwargs.get("cond_dim", 0)
+        # TODO(ilan-gold): What are reasonable defaults here?
+        neural_vf = VelocityField(
+            output_dims=[*self._neural_kwargs.get("velocity_field_output_dims", []), target_dim],
+            condition_dims=(
+                self._neural_kwargs.get("velocity_field_condition_dims", [source_dim + condition_dim])
+                if is_conditional
+                else None
+            ),
+            hidden_dims=self._neural_kwargs.get("velocity_field_hidden_dims", [512, 512, 512]),
```
MUCDK, line 611: default: [1024, 1024, 1024]

src/moscot/backends/ott/solver.py
```diff
@@ -457,3 +500,199 @@ def _call_kwargs(cls) -> Tuple[Set[str], Set[str]]:
...
+        )
+        time_sampler = self._neural_kwargs.get("time_sampler", uniform_sampler)
+        optimizer = self._neural_kwargs.get("optimizer", optax.adam(learning_rate=1e-4))
+        self._solver = genot.GENOT(
+            vf=neural_vf,
+            flow=self._neural_kwargs.get(
+                "flow",
+                dynamics.ConstantNoiseFlow(0.1),
+            ),
+            data_match_fn=functools.partial(data_match_fn, typ="lin", **data_match_fn_kwargs),
```
MUCDK, line 630: can we make `data_match_fn` of type `Optional[Callable[[Optional[ArrayLike], Optional[ArrayLike], Optional[ArrayLike], Optional[ArrayLike]], ArrayLike]]`
so the user can pass their own `data_match_fn`

src/moscot/backends/ott/output.py
```diff
@@ -238,6 +241,227 @@ def _ones(self, n: int) -> ArrayLike:  # noqa: D102
...
+        if forward:
+            if save_transport_matrix:
+                self._transport_matrix = tm
+        else:
+            tm = tm.T
+            if save_transport_matrix:
+                self._inverse_transport_matrix = tm
+        return tm
+
+    def project_transport_matrix(  # type:ignore[override]
```
MUCDK, line 310: I know this is my bad as you inherited. But can we rename this to `project_to_transport_matrix`

src/moscot/problems/generic/_mixins.py
```diff
@@ -256,3 +272,298 @@ def batch_key(self: GenericAnalysisMixinProtocol[K, B], key: Optional[str]) -> N
...
+        return_all: bool = False,
+        return_data: bool = False,
+        **kwargs: Any,
+    ) -> Optional[ApplyOutput_t[K]]:
+        """
+        Push cells.
+
+        Parameters
+        ----------
+        %(source)s
```
MUCDK, line 481: I think we removed docrep

MUCDK (changes requested):
Thanks @ilan-gold .
Maybe let's do the few changes I suggested, and then next week I'll go over it with @selmanozleyen as well.

ilan-gold:
ok @MUCDK !

src/moscot/backends/ott/solver.py
ilan-gold: How would this work given the fact that the default should take arguments?  Allow both kwargs for the default and then a user-defined option?  Make the default ones not take any arguments?

MUCDK: Let's expose `epsilon`,  `tau_a` and  `tau_b`, `scale_cost`, and `cost_fn`  in `data_match_fn` and the actual `solve` function, and then not provide any keyword arguments for it. 

And if `data_match_fn` is given, we say it ignores these parameters. What do you think?

selmanozleyen:
@MUCDK do you think GENOT code in ottjax is final? My version in my thesis project is very different from the one in ottjax . If you like maybe we can first discuss possible changes there. Depends on what you want to prioritize.

MUCDK:
> @MUCDK do you think GENOT code in ottjax is final? My version in my thesis project is very different from the one in ottjax . If you like maybe we can first discuss possible changes there. Depends on what you want to prioritize.

Yeah it's not super final but I wouuld prefer to have the merges more incremental as we have been workin gon this PR for almost 2 years.

Let's talk next week in person.

src/moscot/backends/ott/solver.py
ilan-gold: Sounds good!

ilan-gold: I think we'll need to expose them as before then, so no change.  `tau_a` could be used i.e., for unbalancedness.  So I'll just override use of the default function if provided.

tests/problems/generic/test_gap_problems.py
selmanozleyen, line 1: Is this file left empty intentionally?

src/moscot/problems/time/__init__.py
selmanozleyen, line 1: I know this might be nitpicking but I am just curious if there is a reason for this change

src/moscot/problems/generic/_mixins.py
```diff
@@ -127,9 +127,9 @@ def push(
         Parameters
         ----------
         source
-            Source key in :attr:`solutions`.
+            Source key in `solutions`.
```
selmanozleyen, line 130: Is there a reason for this change?

selmanozleyen (commented):
Thanks @ilan-gold I made some comments. If it passes the relevant tests and docs seem good that's enough for me. But one very important thing is the conflict in the files.

docs/conf.py
```diff
@@ -63,12 +63,24 @@
...
 nitpick_ignore = [
     ("py:class", "numpy.float64"),
+    # see: https://github.com/numpy/numpydoc/issues/275
+    ("py:class", "None.  Remove all items from D."),
+    ("py:class", "a set-like object providing a view on D's items"),
+    ("py:class", "a set-like object providing a view on D's keys"),
+    ("py:class", "v, remove specified key and return the corresponding value."),  # noqa: E501
+    ("py:class", "None.  Update D from dict/iterable E and F."),
+    ("py:class", "an object providing a view on D's values"),
+    ("py:class", "a shallow copy of D"),
```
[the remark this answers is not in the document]
giovp: I think these are warnings from the RTD that should be ignored base on the static class rendering file. But they are indeed too many I'm not sure if it's wanted.

MUCDK:
Greatl, let's wait for @ilan-gold 's response, and then merge.

docs/conf.py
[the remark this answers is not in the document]
ilan-gold: Not sure, what to do here.  We can't import from `MutableMapping` for `DistributionCollection` without re-implementing, which seems like a waste.  The other option would be to remove the `DistributionCollection` class since it is used only for `repr`/cleanness of API.  @MUCDK this code was pulled from your starting point, so perhaps you can chime in.

src/moscot/problems/generic/_mixins.py
ilan-gold: This was an error, kind of.  The mixin has no `solutions`, only the problem itself does.

docs/conf.py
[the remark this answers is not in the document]
MUCDK: I'd be fine with keeping it as it is if it's not too much of an issue. @selmanozleyen wdyt?

tests/problems/generic/test_gap_problems.py
selmanozleyen: maybe a TODO comment instead of leaving empty might be better

MUCDK: we can delete this file

docs/conf.py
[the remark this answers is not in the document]
selmanozleyen: If typing.Dict doesn't work as well I guess using those ignores is a better solution than re-implementation.

Issue #668, opened 2024-03-13
`set_x` adapt assertion


MUCDK:
`pd.testing.assert_series_equal(self.adata_src.[obs_names.to](http://obs_names.to/)_series(), [data.columns.to](http://data.columns.to/)_series()) `

requires me to do

```
early_dist.index.name = 'cell'
late_dist.index.name = 'cell'
early_dist.columns.name = 'cell'
late_dist.columns.name = 'cell'
```
which is a bit on an overkill I think

PR #669, opened 2024-03-13
Don't assert for index/column names when comparing series (e.g. in `set_x`, `set_y`, and `set_x`)

selmanozleyen:
Fixes https://github.com/theislab/moscot/issues/668. To standardize series assertions I used an utility function in all cases where `pd.testing.assert_equal_series` is used.


giovp (approved):

MUCDK (approved):
LGTM. thanks

Issue #667, opened 2024-03-13
add tests that marginals are passed correctly


MUCDK:
This can be done by checking that the prior and posterior marginals are similar up to an error tolerance whenever we solve a balanced problem.

PR #670, opened 2024-03-13
Added tests on passing marginals to `prepare`

selmanozleyen:
Fixes https://github.com/theislab/moscot/issues/667. Added some tests that check if marginals are set.


MUCDK (approved):
Thanks Selman!

PR #677, opened 2024-03-20
Add Sparse Geodesic Cost Support

selmanozleyen:
Here are the steps I planned. I first want to ensure the tests work with this version and I am currently on that step.
- [x] make sure it works on the new commit of ottjax (here: https://github.com/ott-jax/ott/commit/51a658aaa026528a3b9702e3d25c8f7af598cef0)
- [x] don't densify sparse arrays for geodesic costs
- [x] adjust and add tests 

In order to comply to new version:
- [x] created new solution files because loaded objects in test_regression_testing had corrupted format.
- [ ] https://github.com/theislab/moscot/issues/678
- [x] https://github.com/ott-jax/ott/issues/509
- [x] https://github.com/theislab/moscot/pull/686
- [x] https://github.com/theislab/moscot/pull/688 (already included in this PR)
ping: @MUCDK , @giovp 
- [ ] https://github.com/ott-jax/ott/issues/519


giovp:
- try increase the taus for the failing tests for FGW/GW problems
- default to None the n iterations

selmanozleyen:
I tried all of your suggestions and it only worked when I set the tau's to be 1.0 which is balanced. I checked again and it fails after this https://github.com/ott-jax/ott/commit/41906a2a1ade19aa154189fabd7c159a160c9bf3 commit specifically. I created a PR separately for the setting the defaults to None https://github.com/theislab/moscot/pull/686

giovp:
ok then i think we need to raise this to @michalk8 because it is not expected

MUCDK:
agree, can you maybe try to run the example without moscot, but directly with ott-jax @selmanozleyen ? This might help @michalk8

selmanozleyen:
> agree, can you maybe try to run the example without moscot, but directly with ott-jax @selmanozleyen ? This might help @michalk8

@MUCDK ok I created it here https://github.com/ott-jax/ott/issues/519

src/moscot/backends/ott/_utils.py
```diff
@@ -67,6 +68,25 @@ def alpha_to_fused_penalty(alpha: float) -> float:
     return (1 - alpha) / alpha
 
 
+def densify(arr: ArrayLike) -> jax.Array:
+    """If the input is sparse, convert it t dense.
```
MUCDK, line 72: """If the input is sparse, convert it to dense.

MUCDK (changes requested):
Can you please also update the tests with geodesic costs for other problems, e.g. https://github.com/theislab/moscot/blob/6cc22154caa96c30d2319b7268aa1e5d0c482dd2/tests/problems/time/test_temporal_problem.py#L252

Also, accordingly, please update https://github.com/theislab/moscot/blob/6cc22154caa96c30d2319b7268aa1e5d0c482dd2/src/moscot/utils/tagged_array.py#L106

PR #686, opened 2024-04-09
Set defaults of `min_iterations` and `max_iterations` to None

selmanozleyen:
As discussed in https://github.com/theislab/moscot/pull/677#issuecomment-2044341517 since it's hard to catch up with the changes of ottjax defaults we expose these arguments but remove them if not specified.


src/moscot/base/problems/compound_problem.py
```diff
@@ -313,6 +313,11 @@ def solve(
         problems = self._problem_manager.get_problems(stage=stage)
 
         logger.info(f"Solving `{len(problems)}` problems")
+        # expose min/max iterations to the user but remove them if they are None
+        if "min_iterations" in kwargs and kwargs["min_iterations"] is None:
```
giovp, line 317: this could be kwargs.get(min_iterations) and return none same as below

giovp (approved):
minor thing for clarity

MUCDK (approved):

src/moscot/base/problems/compound_problem.py
selmanozleyen: true, thanks solved

selmanozleyen: I did what you said but thinking again if kwargs.get(min_iterations) is None it could be either because it is set to None or key doesn't exist. But the line I wrote was for the case when the key existed but it was set to None. So if I use kwargs.get version in the future it will throw a key error when min_iterations isn't specified at all by the subclasses. So I think I will revert that commit

Issue #678, opened 2024-03-20
No convergence with new commit of ottjax in LRGW solvers unless inner_iterations=10 is set


selmanozleyen:
I am working on [this](https://github.com/theislab/moscot/pull/677) and I am solving problems with the new ottjax version. You can also see the tests that fail from the CI https://github.com/theislab/moscot/actions/runs/8361086332/job/22888306575#step:5:1934.

The code that fails would look something like this:
```python
ap = (
    AlignmentProblem(adata=adata_space_rotate)
    .prepare(batch_key="batch")
    .solve(epsilon=epsilon, alpha=alpha, rank=rank, initializer=initializer)
)

for prob_key in ap:
    assert ap[prob_key].solution.rank == rank
    assert ap[prob_key].solution.converged
```
but I noticed if I set `inner_iterations=10` it converges
```python
ap = (
    AlignmentProblem(adata=adata_space_rotate)
    .prepare(batch_key="batch")
    .solve(epsilon=epsilon, alpha=alpha, rank=rank, initializer=initializer, inner_iterations=10)
)

for prob_key in ap:
    assert ap[prob_key].solution.rank == rank
    assert ap[prob_key].solution.converged
```
The new versions of LRGW solvers in ottjax handle `inner_iterations` differently. This is one example test that fails  https://github.com/theislab/moscot/blob/a5187c037f7137de7324348de886dc4ee7234000/tests/problems/space/test_alignment_problem.py#L81

giovp:
hi @selmanozleyen , yes, you probably noticed this https://github.com/ott-jax/ott/commit/036add26e4f4391479905e71df910f691f94478c , we might have to update the default to mirror the ott-jax default

selmanozleyen:
I think it mirrors the default but the problem is it doesn't converge when it's a higher number like 2000. It does when it is 10. Idk if this is expected

selmanozleyen:
Ok, I will set the inner iterations to 10 as in the tests of ott-jax. Defaults are already passing.

PR #688, opened 2024-04-16
Fixing ottjax 0.4.6 Errors

selmanozleyen:
errors also faced in #677
- [x] test import error, same as this commit from another PR: https://github.com/theislab/moscot/pull/677/commits/fd407df652a1a41f81196927e784a28f58c26b4f
- [x] resave the alignment and mapping pickle file to fix the regression test, this commit https://github.com/theislab/moscot/pull/677/commits/a2589dab71cbfffee53138d7242437397eca992b
- [x] added https://github.com/theislab/moscot/issues/678
- [x] ignore failing test see https://github.com/ott-jax/ott/issues/519


giovp:
- [ ] resave the alignment and mapping pickle file to fix the regression test

giovp:
hi @selmanozleyen thank you, it looks good but could you please skip instead of modifying iterations for the other two that needs to be fixed? so we can keep track of them better, after that, it can be merged!

selmanozleyen:
@giovp ok done

Issue #596, opened 2023-08-18
prepare method does not warn about unknown parameters


Marius1311:
I was just wondering why `problem.prepare` recomputes a PCA, even though I pre-computed one and passed `join_attr='X_pca'` until I realized that I had a typo in there. The `prepare` method does not complain about this, which can lead to unexpected behaviour.

MUCDK:
Hi Marius,

Thanks for opening this issue.

I think we should discuss whether we want to throw an error for unrecognized arguments. We do this in the `solve` methods, as we can follow a clear logic there, i.e. testing whether an argument is an attribute of the solver class.

In `prepare` we don't have a logic like this, so we'd have to hard code the arguments we check for. 

@michalk8 , do you have an opinion on this?

giovp:
we discussed this extensively in the past and I still think it's a pretty bad and unique ux design, wrong arguments should always errors, I understand that we had other priorities but it's something we'll have to figure out sooner than later

Marius1311:
Thanks for your reply @MUCDK - I fully agree with @giovp, not throwing error for unrecognised arguments can get really confusing.

selmanozleyen:
closed with https://github.com/theislab/moscot/pull/696

Marius1311:
amazing, this is great! I think this was very important and will prevent many future mistakes in practical applications.

PR #696, opened 2024-05-21
Refactor: Handling Unrecognized Arguments for Prepare

selmanozleyen:
These are the things I want to do first.  related: #596

From reading through ottjax only the `__call__` functions on solvers absorb kwargs that aren't recognized. And they are because either they are given to create_geometry methods or create_initializer methods:

- https://github.com/ott-jax/ott/blob/458e265a1282d417f87bc0e07a5b6b77ffe9ae50/src/ott/initializers/quadratic/initializers.py#L134-L155
- https://github.com/ott-jax/ott/blob/458e265a1282d417f87bc0e07a5b6b77ffe9ae50/src/ott/initializers/linear/initializers_lr.py#L232-L243

So this is kind of good news because from what I read from moscot, call_fn arguments are filtered before here:
https://github.com/theislab/moscot/blob/e21677e6fbf1e5484b1c69a8ba55587dbf40f009/src/moscot/base/solver.py#L145-L164

Overall this means:
- We can fix the`prepare` methods absorbing kwargs ourselves in moscot, by simply removing the usage of kwargs in concrete classes, so I will dedicate this PR to it
- But for `solve` we will need to refactor things from ottjax. It won't be too hard and I already spotted where they are being ignored. If we refactor ottjax we might not need any changes in moscot since the error will be thrown from there.

Related issues (will be closed after merge):
- https://github.com/theislab/moscot/issues/698
- https://github.com/theislab/moscot/issues/699
- https://github.com/theislab/moscot/issues/700
- https://github.com/theislab/moscot/issues/701

I separated some issues from this PR so that we can work on them later:
- https://github.com/theislab/moscot/issues/716 : I did this one because when inheriting docstring GW removed the arguments for linear, and this lead to duplicate docstrings
- https://github.com/theislab/moscot/issues/715 : Because although in terms of software, current PR makes sense, there might be some cases where a parent classes argument is ignored/fixed but is not yet discovered. I am not confident about this as I am not completely familiar with the Biology use cases of all classes.
- https://github.com/theislab/moscot/issues/721

Now that the tests pass these are the things to do:
- [x] clean up code (most likely requires refactoring of `handle_joint_attr`)
- [x] update docstrings
- [x] clean the TODO's and ensure that everyone agrees on the new docs and docstrings
- [x] write tests to ensure it fails (not sure if tests are necessary at this point since the checks are static now, will probably figure out cases to test when working on https://github.com/theislab/moscot/issues/715)


src/moscot/problems/space/_alignment.py
```diff
@@ -128,23 +132,56 @@ def prepare(
...
+                raise ValueError("Cannot normalize spatial coordinates when `x_callback` and `y_callback` are set.")
+                warnings.warn(
+                    "Ignoring `normalize_spatial` as `x_callback` and `y_callback` are set.",
+                    UserWarning,
+                    stacklevel=2,  # TODO: check these later
+                )
+        # TODO: what if one is spatial norm?
+        if x_callback == "spatial-norm" and y_callback == "spatial-norm":
+            x = y = {}
+        xy, xy_callback, xy_callback_kwargs = handle_joint_attr(joint_attr, xy_callback, xy_callback_kwargs)
```
giovp, line 156: > `# TODO: what if one is spatial norm?`

it must be in either both or none, as there is no use case for only one, hence they are set both by `normalize_spatial` arg

also this apply only for the alignment, as in the mapping it can be only in one

giovp (approved):
fantastic work @selmanozleyen !

src/moscot/problems/space/_alignment.py
selmanozleyen: so should I throw an error when only one is given?

MUCDK: also for spatiotemporal. 
I think a warning would be good

MUCDK (approved):
Really great work Selman! Thanks so much!

src/moscot/base/problems/birth_death.py
```diff
@@ -159,14 +163,53 @@ class BirthDeathProblem(BirthDeathMixin, OTProblem):
         Keyword arguments for :class:`~moscot.base.problems.OTProblem`.
     """  # noqa: D205
 
+    def prepare(
```
MUCDK, line 166: remove the prepare method and pass it via `marginal_kwargs`

selmanozleyen:
@MUCDK  the problem is even though 3.10 passes 3.9 now fails. The results change between these versions

MUCDK:
Ok can we skip the test for python 3.9?

selmanozleyen:
Sure will update the tests in a new PR then

PR #697, opened 2024-05-22
center scale for spatial normalization callback

giovp:


MUCDK (approved):

giovp:
I'll modify the marginals for the unbalanced fgw case with marginals != 1

PR #713, opened 2024-06-12
Fix: CI examples not throwing error from some commands

selmanozleyen:
Here is an example case where it throws error but the CI still doesn't fail. This PR will fix that issue

image

- The script run is now a separate file
- I also added tutorials folder to the list of notebooks to execute
- added the dependencies to add the tutorials folder

It is ready for review. Once the notebooks are updated it is ready for merge imo


pyproject.toml
```diff
@@ -302,12 +302,13 @@ deps =
     jupytext
     nbconvert
     leidenalg
+    squidpy
+    pot
```
MUCDK, line 306: why do we need this?

MUCDK:
Amazing, thanks @selmanozleyen ! @ArinaDanilina , please update the notebooks accordingly!

pyproject.toml
giovp: exactly, installation of squidpy is also in the [spatial] optional requirements

giovp:
I think the tutorial folder is a bit excessive, and examples should be fine?

pyproject.toml
selmanozleyen: For the tutorials folder. If we don't need to check them then I can remove.

MUCDK:
I think tutorials is too much, especially once we replace them with larger data. Then CI would just take for too long. Let's rather have them checked manually regularly

MUCDK (approved):
LGTM, thanks!

PR #714, opened 2024-06-13
update notebooks

ArinaDanilina:


MUCDK (approved):
only if tests pass obv.

selmanozleyen:
Hi I want to note that although the tests passes here it doesnt on the other PR here https://github.com/theislab/moscot/actions/runs/9513260069/job/26222972253#step:5:153 It's just one notebook remaining. Writing so that it won't also fail in the future.

PR #722, opened 2024-07-02
Feat/lin spatial map

Marius1311:
This introduces a `LinearSpatialMapping` problem which is mean for mapping in spatially-aware latent spaces (think about e.g. NicheCompass latent spaces). Thus, spatial information is already contained in latent dimensions and does not need to be supplied externally via a quadratic term. This reduces time complexity, memory complexity, and the number of tunable parameters. 

The current implementation is naive, I just copied the `SpatialMapping` problem and adapted the `prepare` and `solve` methods. What would be a better abstraction in your opinion @giovp? We could do it like you @MUCDK did for the Temporal/Lineage problems, i.e. have the quadratic problem inherit from the linear one, or we could also do it the other way round. Do you prefer any of these possibilities? Or something else?

This PR also introduces a `groupby` parameter for gene correlation computation, which is useful when you're interested in correlation over specific groups, like clusters etc. 

I'll add tests once I hear your general thoughts on this - I'm happy to adapt this to your requirements.

### DIFF
--- a/src/moscot/problems/space/__init__.py
+++ b/src/moscot/problems/space/__init__.py
@@ -1,10 +1,12 @@
 from moscot.problems.space._alignment import AlignmentProblem
+from moscot.problems.space._linear_mapping import LinearMappingProblem
 from moscot.problems.space._mapping import MappingProblem
 from moscot.problems.space._mixins import SpatialAlignmentMixin, SpatialMappingMixin
 
 __all__ = [
     "AlignmentProblem",
     "MappingProblem",
+    "LinearMappingProblem",
     "SpatialAlignmentMixin",
     "SpatialMappingMixin",
 ]
--- a/src/moscot/problems/space/_linear_mapping.py
+++ b/src/moscot/problems/space/_linear_mapping.py
@@ -1,0 +1,320 @@
+import types
+from typing import Any, Literal, Mapping, Optional, Sequence, Tuple, Type, Union
+
+from anndata import AnnData
+
+from moscot import _constants
+from moscot._types import (  # QuadInitializer_t,
+    ArrayLike,
+    CostKwargs_t,
+    OttCostFn_t,
+    Policy_t,
+    ProblemStage_t,
+    ScaleCost_t,
+    SinkhornInitializer_t,
+)
+from moscot.base.problems.compound_problem import B, CompoundProblem, K
+from moscot.base.problems.problem import OTProblem
+from moscot.problems._utils import handle_cost, handle_joint_attr
+from moscot.problems.space._mixins import SpatialMappingMixin
+from moscot.utils.subset_policy import DummyPolicy, ExternalStarPolicy
+
+__all__ = ["LinearMappingProblem"]
+
+
+class LinearMappingProblem(SpatialMappingMixin[K, OTProblem], CompoundProblem[K, OTProblem]):
+    """Class for mapping single cell omics data onto spatial data, based on :cite:`nitzan:19`.
+
+    Parameters
+    ----------
+    adata_sc
+        Annotated data object containing the single-cell data.
+    adata_sp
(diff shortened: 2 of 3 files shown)


MUCDK:
Thanks @Marius1311 !
Let's go step by step.
To what extent is this necessary given that we have the `SinkhornProblem`? How common is this use case?

giovp:
agree with @MUCDK , what about making the mapping problem be arbitrarily also used in pure linear setting? I like the groupby addition! could we maybe have that in as separate PR?

Marius1311:
> Thanks @Marius1311 ! Let's go step by step. To what extent is this necessary given that we have the `SinkhornProblem`? How common is this use case?

It's necessary as you don't get the relevant downstream analysis with the SinkhornProblem. I don't think this usage is very common yet, but I need this as part of a method I'm currently developing. 

So far, I solved a generalized Sinkhorn problem and then manually transferred the solution into a SpatialMapping Problem for the downstream analysis. But that's not very convenient, and I'm not sure how efficient this is, given that the solution to a LinearProblem can be queried batch-wise, so you have a much smaller memory footprint for downstream analysis, like imputation or correlation computation. If I'm understanding things correctly, then batch-wise evaluation of the mapping is not possible for a (full-rank) quadratic problem.

Keep in mind that my indented use-case here is full-rank, not low-rank.

Marius1311:
> agree with @MUCDK , what about making the mapping problem be arbitrarily also used in pure linear setting? I like the groupby addition! could we maybe have that in as separate PR?

Re 1, I agree, that would be another possibility, although I think it's cleaner to have two separate problems as you might want to have different default values etc. Also, I think it would be more user-friendly to have two separate problems as users of the Linear Problem would only be interested in a subset of the parameters required for the Quadratic problem. 

Re 2, sure, I can move the groupby addition into a separate PR.

giovp:
Hi @Marius1311 we discussed a bit with @MUCDK and @selmanozleyen and we think that to turn the MappingProblem into a LinearMappingProblem it would actually only require to set alpha=0 at the MappingProblem level and then re-set the x and y to None. This would automatically solve a linear problem and not a fused problem. Because it is a very simple change to be made, at the biological problem level (MappingProblem), then I think it would not require an entire new problem but just a different way to handle the arguments. 

The behaviour could simply be the following. If `alpha=0` then it would turn x and y to `None` and raise a `logg.info` specifying that the geometries for the quadratic term are not being used. 

I you agree, could you kindly change the code to just implement this behavior in the existing mapping problem, and then submit a separate PR for he groupby? Let us know if anything is unclear, thank you!

MUCDK:
> Hi @Marius1311 we discussed a bit with @MUCDK and @selmanozleyen and we think that to turn the MappingProblem into a LinearMappingProblem it would actually only require to set alpha=0 at the MappingProblem level and then re-set the x and y to None. This would automatically solve a linear problem and not a fused problem. Because it is a very simple change to be made, at the biological problem level (MappingProblem), then I think it would not require an entire new problem but just a different way to handle the arguments.
> 
> The behaviour could simply be the following. If `alpha=0` then it would turn x and y to `None` and raise a `logg.info` specifying that the geometries for the quadratic term are not being used.
> 
> I you agree, could you kindly change the code to just implement this behavior in the existing mapping problem, and then submit a separate PR for he groupby? Let us know if anything is unclear, thank you!

see e.g. https://github.com/theislab/moscot/blob/e5a0e0db069dd2e12bab7f3cdb8c5041ea0f2f31/src/moscot/base/problems/problem.py#L298

Marius1311:
That sounds good, will try to do that and let you know if I encounter any difficulties! Thanks!

Marius1311:
I'm closing this in favour of https://github.com/theislab/moscot/pull/732, where I included your suggestions @MUCDK @giovp

PR #723, opened 2024-07-03
Fix: test_compute_interpolated_distance_regression and numerical column error

selmanozleyen:
Hi @MUCDK @giovp ,

When running generate_gt_temporal_data I encountered an error because adata[day] wasn't categorical but was a float. So in the code I cast it as a category. Not sure why this happens now. But in the end I created the file and it should help the tests pass. But as I said I am not sure what I should do for the temporal key.

After this PR, ".A" PR will also be ready to merge. Then the "prepare" PR


MUCDK:
That’s fine. I think when we generated the data we didn’t have the restriction for the temporal key to be of categorical type. But it’s true that now we do check for it. So all good! Thanks for this fix

MUCDK (approved):

PR #725, opened 2024-07-03
Skip Regression Tests For Versions Lower Than Python 3.9 and Remove Elastic Cost

selmanozleyen:
This PR fixes tests for the CI.
skipping regression tests depending on the version and removing elastic costs.


MUCDK (approved):
Thanks!

selmanozleyen:
@MUCDK And now it fails because ElasticL1 was removed from ottjax 2 hours ago.

MUCDK:
remove the tests for Elastic costs, and open an issue to investigate later. I saw that Mike restructured quite a. bit. Elastic costs are not super crucial , so fine to have them not for the moment.

Thanks!

Issue #703, opened 2024-05-28
check the convergence flag


MUCDK:
in the LR case the flag is raised despite convergence.

PR #728, opened 2024-07-03
Remove skip markers for balanced since LR criterion is fixed

selmanozleyen:
After this https://github.com/ott-jax/ott/pull/547#event-13382808561 we can finally expect the tests to pass again. So the skip markers are removed.

Closes:
- https://github.com/theislab/moscot/issues/678
- https://github.com/theislab/moscot/issues/703


giovp:
@selmanozleyen I have done that also here just FYI https://github.com/theislab/moscot/pull/697/files

selmanozleyen:
Ah ok then I am closing this

PR #731, opened 2024-07-09
Introduce a `groupby` attribute for correlation computation

Marius1311:
This introduces a `groupby` parameter for `mapping_problem.correlate()` to evaluate Pearson/Spearmann correlation over groups of cells. I've also adapted tests slightly to cover this case.

### DIFF
--- a/src/moscot/problems/space/_mixins.py
+++ b/src/moscot/problems/space/_mixins.py
@@ -454,7 +454,7 @@ class SpatialMappingMixin(AnalysisMixin[K, B]):
             # initialize a dict of group masks
             if groupby:
                 groups = adata_sp.obs[groupby].cat.categories
-                group_masks = {group: adata_sp.obs[groupby] == group for group in groups}
+                group_masks = {group: (adata_sp.obs[groupby]).values == group for group in groups}
                 corrs[key] = {}
             else:
                 group_masks = {"all": np.ones(adata_sp.shape[0], dtype=bool)}
@@ -472,6 +472,7 @@ class SpatialMappingMixin(AnalysisMixin[K, B]):
                     logger.debug(f"Skipping `group={group}` as it contains less then 2 samples.")
                     continue
 
+                # import pdb; pdb.set_trace()
                 corr_val = [
                     corr(gexp_pred_sp[group_mask, gi], gexp_sp[group_mask, gi])[0] for gi, _ in enumerate(var_sc)
                 ]
--- a/src/moscot/problems/space/_mixins.py
+++ b/src/moscot/problems/space/_mixins.py
@@ -414,10 +414,8 @@ class SpatialMappingMixin(AnalysisMixin[K, B]):
             - ``'pearson'`` - `Pearson correlation <https://en.wikipedia.org/wiki/Pearson_correlation_coefficient>`_.
             - ``'spearman'`` - `Spearman's rank correlation
                 <https://en.wikipedia.org/wiki/Spearman%27s_rank_correlation_coefficient>`_.
-
         groupby
             Optional `obs` field in `AnnData` to compute correlations over categorical groups.
-
         device
             Device where to transfer the solutions, see :meth:`~moscot.base.output.BaseSolverOutput.to`.
 
@@ -440,7 +438,7 @@ class SpatialMappingMixin(AnalysisMixin[K, B]):
         if sp.issparse(gexp_sc):
             gexp_sc = gexp_sc.toarray()
 
-        corrs = {}  # type: ignore
+        corrs: Union[Dict[Tuple[K, K], Dict[Any, pd.Series]], Dict[Tuple[K, K], pd.Series]] = {}
         for key, val in self.solutions.items():
 
             # create mask corresponding to the current batch of spatial data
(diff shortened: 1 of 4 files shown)


src/moscot/problems/space/_mixins.py
```diff
@@ -396,7 +396,9 @@ def correlate(  # type: ignore[misc]
         self: SpatialMappingMixinProtocol[K, B],
         var_names: Optional[Sequence[str]] = None,
         corr_method: Literal["pearson", "spearman"] = "pearson",
-    ) -> Mapping[Tuple[K, K], pd.Series]:
+        device: Optional[Device_t] = None,
+        groupby: Optional[str] = None,
```
giovp, line 400: what about calling it `groups`? My understanding is that this is used to compute correlation between groups, see lines 457, 458. I think `groups` is more fitting and to some extent mirrors the usage of the argument in other scverse packages.

src/moscot/problems/space/_mixins.py
```diff
@@ -411,7 +413,13 @@ def correlate(  # type: ignore[misc]
 
             - ``'pearson'`` - `Pearson correlation <https://en.wikipedia.org/wiki/Pearson_correlation_coefficient>`_.
             - ``'spearman'`` - `Spearman's rank correlation
-              <https://en.wikipedia.org/wiki/Spearman%27s_rank_correlation_coefficient>`_.
+                <https://en.wikipedia.org/wiki/Spearman%27s_rank_correlation_coefficient>`_.
+
+        groupby
+            Optional `obs` field in `AnnData` to compute correlations over categorical groups.
+
```
giovp, line 420: remove the line space between arguments

src/moscot/problems/space/_mixins.py
```diff
@@ -432,19 +440,46 @@ def correlate(  # type: ignore[misc]
         if sp.issparse(gexp_sc):
             gexp_sc = gexp_sc.toarray()
 
-        corrs = {}
+        corrs = {}  # type: ignore
```
giovp, line 443:
```suggestion
        corrs: dict[str,Union[pd.Series,dict[str,pd.Series]] = {}
```

giovp: this might work. Would try to type it correctly

src/moscot/problems/space/_mixins.py
```diff
@@ -432,19 +440,46 @@ def correlate(  # type: ignore[misc]
...
+            # predict spatial feature expression
+            gexp_pred_sp = val.to(device=device).pull(gexp_sc, scale_by_marginals=True)
+
+            # loop over groups and compute correlations
+            for group, group_mask in group_masks.items():
+                if np.sum(group_mask) < 2:
+                    logger.debug(f"Skipping `group={group}` as it contains less then 2 samples.")
+                    continue
+
+                # import pdb; pdb.set_trace()
```
giovp, line 475: remove

giovp (changes requested):
looks good! Test also looks good. Minor comments. Thanks for the PR!

src/moscot/problems/space/_mixins.py
Marius1311: I actually prefer `groupby` - you still get a vector of the same length, i.e. length `n_genes`. The difference to before it that when you compute correlation, you only take into account the cells in the current group. In other words, you now get one such vector (containing correlations) per problem and per group. 

I tried to follow scanpy here a bit, see e.g. the violin plotting function, which uses `grouopby` in a similar fashion. https://scanpy.readthedocs.io/en/stable/generated/scanpy.pl.violin.html

On the other hand, in scanpy's scatter plot function, they use a `groups` attribute to restrict visualization to certain groups, which is not really what we are doing here: https://scanpy.readthedocs.io/en/stable/generated/scanpy.pl.scatter.html

src/moscot/problems/space/_mixins.py
Marius1311: done.

src/moscot/problems/space/_mixins.py
Marius1311: done.

src/moscot/problems/space/_mixins.py
Marius1311: done.

Marius1311:
Thanks for the review @giovp, incorporated all your suggestions, except for the `groupby` -> `groups` change, see comment above.

selmanozleyen (approved):
Looks good

Marius1311:
Great, thanks @selmanozleyen @giovp, merging this.

selmanozleyen:
@giovp  I think the problem with linting is here for sure. I thought maybe it's a link checking error because of connection but it fails only after this branch (I rerun the checks on older PR's)

giovp:
yeah I think the problem is something in the docs of the `correlate` method, not a linkcheck

selmanozleyen:
It's from the warning. Since the `-W` flag is set it gives error when warning is issued. I removed the flag and it works locally https://www.sphinx-doc.org/en/master/man/sphinx-build.html

PR #732, opened 2024-07-09
Modify the SpatialMapping problem to enable solving a LinearProblem

Marius1311:
This just checks whether `alpha=0` and in that case sets the quadratic attributes to `None`. 

Let me know what you think, I'm happy to add tests for this.

### DIFF
--- a/src/moscot/backends/ott/_utils.py
+++ b/src/moscot/backends/ott/_utils.py
@@ -63,9 +63,9 @@ def check_shapes(geom_x: geometry.Geometry, geom_y: geometry.Geometry, geom_xy: 
 
 def alpha_to_fused_penalty(alpha: float) -> float:
     """Convert."""
-    if not (0 <= alpha <= 1):
-        raise ValueError(f"Expected `alpha` to be in interval `[0, 1]`, found `{alpha}`.")
-    return (1 - alpha) / alpha if alpha != 0 else 0
+    if not (0 < alpha <= 1):
+        raise ValueError(f"Expected `alpha` to be in interval `(0, 1]`, found `{alpha}`.")
+    return (1 - alpha) / alpha
 
 
 def densify(arr: ArrayLike) -> jax.Array:
--- a/src/moscot/backends/ott/_utils.py
+++ b/src/moscot/backends/ott/_utils.py
@@ -63,9 +63,9 @@ def check_shapes(geom_x: geometry.Geometry, geom_y: geometry.Geometry, geom_xy: 
 
 def alpha_to_fused_penalty(alpha: float) -> float:
     """Convert."""
-    if not (0 < alpha <= 1):
-        raise ValueError(f"Expected `alpha` to be in interval `(0, 1]`, found `{alpha}`.")
-    return (1 - alpha) / alpha
+    if not (0 <= alpha <= 1):
+        raise ValueError(f"Expected `alpha` to be in interval `[0, 1]`, found `{alpha}`.")
+    return (1 - alpha) / alpha if alpha != 0 else 0
 
 
 def densify(arr: ArrayLike) -> jax.Array:
--- a/src/moscot/problems/space/_mapping.py
+++ b/src/moscot/problems/space/_mapping.py
@@ -193,6 +193,9 @@ class MappingProblem(SpatialMappingMixin[K, OTProblem], CompoundProblem[K, OTPro
             x = {}
             y = {}
             logger.info("Preparing a :term:`linear problem`.")
+            if var_names and len(var_names) == 0:
+                raise ValueError("Expected `var_names` to be non-empty for a :term:`linear problem`.")
+
         else:
             raise ValueError(
                 "You either need to set both attr:`spatial_key` and attr:`sc_attr` (for a :term:`quadratic problem`)",
@@ -336,6 +339,8 @@ class MappingProblem(SpatialMappingMixin[K, OTProblem], CompoundProblem[K, OTPro
 
         # convert problem type to linear for alpha=0
         if alpha == 0.0 and self.problem_kind != "linear":
+            if self.filtered_vars and len(self.filtered_vars) == 0:
+                raise ValueError("Expected `var_names` to be non-empty for a :attr:`alpha`=0.")
             logger.info("Ignoring quadratic terms for :attr:`alpha=0`.")
             self._problem_kind = "linear"
             for _, value in self.problems.items():
--- a/src/moscot/problems/space/_mapping.py
+++ b/src/moscot/problems/space/_mapping.py
@@ -334,17 +334,25 @@ class MappingProblem(SpatialMappingMixin[K, OTProblem], CompoundProblem[K, OTPro
             **kwargs,
         }
 
-        if alpha == 0.0 or self.problem_kind == "linear":
-            logger.info("Ignoring quadratic terms for :attr:`alpha=0` and for `'linear'` problems.")
+        # convert problem type to linear for alpha=0
+        if alpha == 0 and self.problem_kind != "linear":
+            logger.info("Ignoring quadratic terms for :attr:`alpha=0`.")
             for _, value in self.problems.items():
                 value._x = None
                 value._y = None
                 value._problem_kind = "linear"
+
+        # prepare solver kwargs, depending on the problem type
+        if self.problem_kind == "linear":
+            if alpha is not None and alpha > 0:
+                logger.warning("Ignoring :attr:`alpha` for `'linear'` problems.")
             if "lse_mode" in linear_solver_kwargs:
                 solve_kwargs["lse_mode"] = linear_solver_kwargs["lse_mode"]
             if "inner_iterations" in linear_solver_kwargs:
                 solve_kwargs["inner_iterations"] = linear_solver_kwargs["inner_iterations"]
         else:
+            if alpha is None:
+                raise ValueError("Expected :attr:`alpha` to be in interval `[0, 1]`, found `None`.")
             solve_kwargs["alpha"] = alpha
             solve_kwargs["linear_solver_kwargs"] = linear_solver_kwargs
 
--- a/src/moscot/problems/space/_mapping.py
+++ b/src/moscot/problems/space/_mapping.py
@@ -76,7 +76,7 @@ class MappingProblem(SpatialMappingMixin[K, OTProblem], CompoundProblem[K, OTPro
         self,
         sc_attr: Optional[Union[str, Mapping[str, Any]]],
         batch_key: Optional[str] = None,
-        spatial_key: Optional[Union[str, Mapping[str, Any]]] = "spatial",
+        spatial_key: Union[str, Mapping[str, Any]] = "spatial",
         var_names: Optional[Sequence[str]] = None,
         normalize_spatial: bool = True,
         joint_attr: Optional[Union[str, Mapping[str, Any]]] = None,
@@ -177,7 +177,7 @@ class MappingProblem(SpatialMappingMixin[K, OTProblem], CompoundProblem[K, OTPro
         - :attr:`problem_kind` - set to ``'quadratic'`` (if both ``spatial_key`` and ``sc_attr`` are passed)
             or ``'linear'`` (if both ``spatial_key`` and ``sc_attr`` are `None`).
         """
-        if spatial_key and sc_attr:
+        if sc_attr:
             x = {"attr": "obsm", "key": spatial_key} if isinstance(spatial_key, str) else spatial_key
             y = {"attr": "obsm", "key": sc_attr} if isinstance(sc_attr, str) else sc_attr
 
@@ -189,19 +189,14 @@ class MappingProblem(SpatialMappingMixin[K, OTProblem], CompoundProblem[K, OTPro
                 x = {}
             self.spatial_key = spatial_key if isinstance(spatial_key, str) else spatial_key["key"]
             logger.info("Preparing a :term:`quadratic problem`.")
-        elif spatial_key is None and sc_attr is None:
+        else:
             x = {}
             y = {}
             logger.info("Preparing a :term:`linear problem`.")
             if var_names and len(var_names) == 0:
                 raise ValueError("Expected `var_names` to be non-empty for a :term:`linear problem`.")
 
-        else:
-            raise ValueError(
-                "You either need to set both attr:`spatial_key` and attr:`sc_attr` (for a :term:`quadratic problem`)",
-                "or none of them (for a term:`linear problem`).",
-            )
-
+        self.spatial_key = spatial_key if isinstance(spatial_key, str) else spatial_key["key"]
         self.batch_key = batch_key
         self.filtered_vars = var_names
 
@@ -239,7 +234,7 @@ class MappingProblem(SpatialMappingMixin[K, OTProblem], CompoundProblem[K, OTPro
 
     def solve(
         self,
-        alpha: Optional[float] = 0.5,
+        alpha: float = 0.5,
         epsilon: float = 1e-2,
         tau_a: float = 1.0,
         tau_b: float = 1.0,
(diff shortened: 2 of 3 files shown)


src/moscot/problems/space/_mapping.py
```diff
@@ -314,11 +318,28 @@ def solve(
             min_iterations=min_iterations,
             max_iterations=max_iterations,
             threshold=threshold,
-            linear_solver_kwargs=linear_solver_kwargs,
             device=device,
             **kwargs,
         )
 
+        if alpha == 0.0:
+            logger.info("Ignoring quadratic terms for `alpha=0`, solving a linear problem.")
```
giovp, line 326:
```suggestion
            logger.info("Ignoring quadratic terms for `alpha=0`, solving a :term:`linear problem`.")
```

src/moscot/backends/ott/_utils.py
```diff
@@ -63,9 +63,9 @@ def check_shapes(geom_x: geometry.Geometry, geom_y: geometry.Geometry, geom_xy:
 
 def alpha_to_fused_penalty(alpha: float) -> float:
     """Convert."""
-    if not (0 < alpha <= 1):
-        raise ValueError(f"Expected `alpha` to be in interval `(0, 1]`, found `{alpha}`.")
-    return (1 - alpha) / alpha
+    if not (0 <= alpha <= 1):
+        raise ValueError(f"Expected `alpha` to be in interval `[0, 1]`, found `{alpha}`.")
+    return (1 - alpha) / alpha if alpha != 0 else 0
```
giovp, line 68: why is this required? this shouldn't be raised if alpha=0 in the mapping problem (as you are correctly removing it from the solver_kwargs)

giovp (changes requested):
quick glance and it looks great! Would you mind adding a test? it could just check that the solution is the correct OTT Output (SinkhornOutput I think it is/was called)

MUCDK (changes requested):
So prepare we leave as it is? i.e. we always have to provide the spatial key?

giovp:
> So prepare we leave as it is? i.e. we always have to provide the spatial key?

mmh good point, it should be possible to allow for `None` there in that case I believe, could you also update the prepare method @Marius1311 thanks!

src/moscot/problems/space/_mapping.py
Marius1311: thanks!

src/moscot/problems/space/_mapping.py
```diff
@@ -261,8 +263,8 @@ def solve(
             Parameter in :math:`(0, 1]` that defines how much :term:`unbalanced <unbalanced OT problem>` is the problem
             on the target :term:`marginals`. If :math:`1`, the problem is :term:`balanced <balanced OT problem>`.
         rank
-            Rank of the :term:`low-rank OT` solver :cite:`scetbon:21b`.
-            If :math:`-1`, full-rank solver :cite:`peyre:2016` is used.
+            Rank of the :term:`low-rank OT` solver :cite:`scetbon:21a,scetbon:21b`.
+            If :math:`-1`, full-rank solver :cite:`cuturi:2013,peyre:2016` is used.
         scale_cost
             How to re-scale the cost matrices. If a :class:`float`, the cost matrices
             will be re-scaled as :math:`\frac{\text{cost}}{\text{scale_cost}}`.
```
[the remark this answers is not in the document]
Marius1311: mh, sorry, I don't know what you mean - in the TemporalProblem, the docstring for the `initializer` says:

```
initializer
            How to initialize the solution. If :obj:`None`, ``'default'`` will be used for a full-rank solver and
            ``'rank2'`` for a low-rank solver.
```
so that's the same - how would you adapt that?

src/moscot/backends/ott/_utils.py
Marius1311: you are right, this is not required, reverting this.

src/moscot/problems/space/_mapping.py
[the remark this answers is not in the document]
Marius1311: ok great, thanks!

Marius1311:
> > So prepare we leave as it is? i.e. we always have to provide the spatial key?
> 
> mmh good point, it should be possible to allow for `None` there in that case I believe, could you also update the prepare method @Marius1311 thanks!

I think I also have to make the `sc_attribute` optional in that case. Otherwise, when not providing a `spatial_key`, I get

```bash
File ~/Projects/moscot-fork/src/moscot/base/problems/problem.py:376, in OTProblem.prepare(self, xy, x, y, a, b, marginal_kwargs)
    [374](https://file+.vscode-resource.vscode-cdn.net/Users/mlange/Projects/spatial_analysis/analysis/experiments_and_tutorials/~/Projects/moscot-fork/src/moscot/base/problems/problem.py:374)     self._problem_kind = "quadratic"
    [375](https://file+.vscode-resource.vscode-cdn.net/Users/mlange/Projects/spatial_analysis/analysis/experiments_and_tutorials/~/Projects/moscot-fork/src/moscot/base/problems/problem.py:375) else:
--> [376](https://file+.vscode-resource.vscode-cdn.net/Users/mlange/Projects/spatial_analysis/analysis/experiments_and_tutorials/~/Projects/moscot-fork/src/moscot/base/problems/problem.py:376)     raise ValueError("Unable to prepare the data. Either only supply `xy=...`, or `x=..., y=...`, or all.")
    [377](https://file+.vscode-resource.vscode-cdn.net/Users/mlange/Projects/spatial_analysis/analysis/experiments_and_tutorials/~/Projects/moscot-fork/src/moscot/base/problems/problem.py:377) # fmt: on
    [378](https://file+.vscode-resource.vscode-cdn.net/Users/mlange/Projects/spatial_analysis/analysis/experiments_and_tutorials/~/Projects/moscot-fork/src/moscot/base/problems/problem.py:378) self._a = self._create_marginals(self.adata_src, data=a, source=True, marginal_kwargs=marginal_kwargs)

ValueError: Unable to prepare the data. Either only supply `xy=...`, or `x=..., y=...`, or all.
```

src/moscot/problems/space/_mapping.py
```diff
@@ -222,15 +239,15 @@ def prepare(
 
     def solve(
         self,
-        alpha: float = 0.5,
+        alpha: Optional[float] = 0.5,
```
selmanozleyen, line 242: Why is alpha optional here? Optional means Union[float, None] but I am not sure we want that. Looking at the code it isn't handled when alpha=None so I am assuming this is a mistake.

src/moscot/problems/space/_mapping.py
```diff
@@ -299,24 +319,51 @@ def solve(
...
+        if alpha == 0.0 and self.problem_kind != "linear":
+            if self.filtered_vars and len(self.filtered_vars) == 0:
+                raise ValueError("Expected `var_names` to be non-empty for a :attr:`alpha`=0.")
+            logger.info("Ignoring quadratic terms for :attr:`alpha=0`.")
+            self._problem_kind = "linear"
+            for _, value in self.problems.items():
+                value._x = None
+                value._y = None
+                value._problem_kind = "linear"
+
```
selmanozleyen, line 350: Isn't this done automatically? I am asking because I am not sure. If not I think it's better to change behaviour of CompoundProblem because setting this later seems a bit hacky

selmanozleyen:
@Marius1311  why can't we just pass solver args until we solve https://github.com/theislab/moscot/issues/720? If incompatible x and y are given shouldn't the parent class already give an error itself?

src/moscot/problems/space/_mapping.py
giovp: mmh I don't think it is, but also I haven't tested, but I thought that the alpha value would check whether the geometries where correct and if not raised error. Here, it makes sure that the alpha and geometries are correct. I think it would be useful to understand what is the condition for raising error in the OT problem where this is passed though.

src/moscot/problems/space/_mapping.py
giovp: I agree! I think we should stick to passing alpha=0 for linear problem @Marius1311

src/moscot/problems/space/_mapping.py
selmanozleyen: I checked and it is handled here https://github.com/theislab/moscot/blob/3fd4544b6aa3a13a91dca4f80a85110a879c66ed/src/moscot/base/problems/problem.py#L371-L376. Even if it wasn't we should've modified it that way because in principle we shouldn't access private attributes from this class.

src/moscot/problems/space/_mapping.py
```diff
@@ -299,24 +319,51 @@ def solve(
...
+            if "inner_iterations" in linear_solver_kwargs:
+                solve_kwargs["inner_iterations"] = linear_solver_kwargs["inner_iterations"]
+        else:
+            if alpha is None:
+                raise ValueError("Expected :attr:`alpha` to be in interval `[0, 1]`, found `None`.")
+            solve_kwargs["alpha"] = alpha
+            solve_kwargs["linear_solver_kwargs"] = linear_solver_kwargs
+
+        return super().solve(  # type: ignore[return-value]
+            **solve_kwargs,
```
selmanozleyen, line 366: I tried and this also works. I'd prefer letting parent classes handle the incompatible arguments with alpha. Some cases of the tests fail since alpha=1.0 and xy is given. It would be also cool to assert that the tests send the desired error message: https://stackoverflow.com/questions/23337471/how-do-i-properly-assert-that-an-exception-gets-raised-in-pytest

```python
        additonal_kwargs = {}
        if alpha is not None and alpha > 0.0:
            additonal_kwargs["alpha"] = alpha
            additonal_kwargs["linear_solver_kwargs"] = linear_solver_kwargs
        else:
            additonal_kwargs.update(linear_solver_kwargs)
        return super().solve(  # type: ignore[return-value]
            epsilon=epsilon,
            tau_a=tau_a,
            tau_b=tau_b,
            rank=rank,
            scale_cost=scale_cost,
            batch_size=batch_size,
            stage=stage,
            initializer=initializer,
            initializer_kwargs=initializer_kwargs,
            jit=jit,
            min_iterations=min_iterations,
            max_iterations=max_iterations,
            threshold=threshold,
            device=device,
            **kwargs,
            **additonal_kwargs,
        )
```

selmanozleyen (commented):
Some notes besides my specific comments I made:
- I wouldn't use or set _problem_kind or other private attributes of parent class. It's also not necessary (see my comments on code snippets please)
- I am not sure about not setting a default for `Optional` arguments. It feels like setting `sc_attr=None` would make sense since it's annotated `Optional`.
- Creating tests when it should fail and asserting the exceptions would be very useful imo

MUCDK:
> Some notes besides the specific comments I made:
> 
> * I wouldn't use or set _problem_kind or other private attributes of parent class. It's also not necessary (see my comments on code snippets please)
> * I am not sure about not setting a default for `Optional` arguments. It feels like setting `sc_attr=None` would make sense since it's annotated `Optional`.
> * Creating tests when it should fail and asserting the exceptions would be very useful imo

Agree, but not on default for default with `Optional`. I think the most standard setting is that you do use the quadratic attributes, hence would keep the defaults

src/moscot/problems/space/_mapping.py
Marius1311: I find it a bit unintuitive that when you prepare a problem without quadratic terms,  you still need to explicitly pass `alpha=0` in the solve method, that's why I thought it should be optional in that case.

src/moscot/problems/space/_mapping.py
Marius1311: Thanks for the feedback @selmanozleyen! There's an important case in which this does not work: when the user specifies both the `sc_attr` and the `spatial_key` in the `prepare` method, but then passes `alpha=0` in `solve`. Here's why that case is important: 

For some downstream analysis functions, the `spatial_key` needs to be set (e.g. `spatial_correspondence`), even though we might not use this data directly for solving the mapping problem. We cannot currently set just the `spatial_key` but not the `sc_attr`, as the backend requires either both or none of them. 

So in that case, we prepare a quadratic problem, we do not modify it or pass `alpha`, and upon calling `solve`, we get `ValueError: Unable to solve a quadratic problem with `alpha = 1` and `xy` supplied.`

If we pass `additonal_kwargs["alpha"] = 0` also in the `else` clause in the `solve` method, we get `ValueError: Expected `alpha` to be in interval `(0, 1]`, found `0`.`. I'm not sure how to go about this. 

I think we need a mechanism to set the `spatial_key`, while still either preparing, or somehow converting, to a linear problem, rather than a quadratic one.

Marius1311: One way to go about this would be to handle everything in the `prepare` method, and to make the problem type entirely dependent on `sc_attr`: 
- if it's specified, we're solving a quadratic problem. `alpha=0` is not allowed in `solve`. 
- if it's `None`, we're solving a linear problem. `alpha` is ignored in `solve` (and we can warn about this). 

This way, the `solve` method cannot change the problem type. The problem type is determined on `prepare`, and it depends entirely on `sc_attr`. I pushed a corresponding suggestion, let me know what you think.

Marius1311:
> > Some notes besides the specific comments I made:
> > 
> > * I wouldn't use or set _problem_kind or other private attributes of parent class. It's also not necessary (see my comments on code snippets please)
> > * I am not sure about not setting a default for `Optional` arguments. It feels like setting `sc_attr=None` would make sense since it's annotated `Optional`.
> > * Creating tests when it should fail and asserting the exceptions would be very useful imo
> 
> Agree, but not on default for default with `Optional`. I think the most standard setting is that you do use the quadratic attributes, hence would keep the defaults

So should we set a default for `sc_attr` or do you want to leave it as `sc_attr: Optional[Union[str, Mapping[str, Any]]],`?

Marius1311:
> Some notes besides the specific comments I made:
> 
> * I wouldn't use or set _problem_kind or other private attributes of parent class. It's also not necessary (see my comments on code snippets please)
> * I am not sure about not setting a default for `Optional` arguments. It feels like setting `sc_attr=None` would make sense since it's annotated `Optional`.
> * Creating tests when it should fail and asserting the exceptions would be very useful imo

Makes sense, thank you! I agree we should provide some default for `sc_attr` if we want to annotate it as `Optional`. I'll add additional tests once we converge on a rationale for determining the problem type, see my comments above.

src/moscot/problems/space/_mapping.py
selmanozleyen: What you suggested seems reasonable. In the future we might also require a case where we want to specify both but still solve it as a linear problem. I think in that case we would've take the problem type as an argument for prepare. But since sc_attr seems to imply the problem type there is no such need. Either case is better than changing the problem type imo.

giovp:
Late on the discussion, I think it makes sense that `sc_attr` can be None, but then if that's the case and `alpha!=0` then this raises an error. Instead, if `sc_attr` is None, but `alpha=0` then it solves correctly. 

On the other way, if `sc_attr` is not None, but `alpha=0` then the solve method should also raise an error. I believe this is the current implementation correct?

src/moscot/problems/space/_mapping.py
Marius1311: makes sense!

Marius1311:
Hi all, I included all of your suggestions. In particular, @giovp, I the code now handles both cases as you expected in your last comment, and I added a test for both cases which checks for the correct exception to be raised. 

Let me know if anything else is required from your sides @selmanozleyen @MUCDK.

src/moscot/problems/space/_mapping.py
```diff
@@ -299,8 +314,17 @@ def solve(
...
         """
+        additonal_kwargs = {}
+        if self.problem_kind == "quadratic":
+            additonal_kwargs["alpha"] = alpha
+            additonal_kwargs["linear_solver_kwargs"] = linear_solver_kwargs
+        else:
+            if alpha != 0:
+                raise ValueError("Expected `alpha` to be 0 for a `linear problem`.")
+            additonal_kwargs.update(linear_solver_kwargs)
+            # logger.warning("Solving a `linear problem`, ignoring `alpha`.")
```
selmanozleyen, line 325: Maybe you forgot to delete this commented line 325

selmanozleyen (approved):
Besides one minor comment it looks good. Thanks! I think testing the corner cases is very useful

src/moscot/problems/space/_mapping.py
Marius1311: ah great, thanks!

Marius1311:
> Besides one minor comment it looks good. Thanks! I think testing the corner cases is very useful

Amazing, thanks for your review @selmanozleyen !

Marius1311:
Plotting tests are failing unrelated to this, so I'm merging.

giovp:
@Marius1311 the tests were related to the plotting, but the docs linting was not. The docs are failing due to this, do you know what could it be 
image

Marius1311:
sorry about that - would have fixed it now, but just saw https://github.com/theislab/moscot/pull/737

giovp:
yeah np we are fixing it there. Btw there are docs errors also in the correlate method, but we'll also just fix everything there np!

Issue #729, opened 2024-07-05
`.impute` consumes too much memory 


Marius1311:
I'm trying to call `problem.impute()` on a solved (linear) spatial mapping problem of dimensions `n_source=17806` (spatial data) by `n_target=13298` (single-cell data) for `n_genes=2039`. This is just a full-rank Sinkhorn problem with `batch_size=None`. 

Under the hood, this evaluates:
```Python
 predictions = [val.to(device=device).pull(gexp_sc, scale_by_marginals=True) for val in self.solutions.values()]
```

The `pull` amounts to a matrix multiplication: `prediction = P @ X` for transport matrix of shape `17806 x 13298` and single-cell GEX matrix `X` of shape `13298 x 2039`. Thus, the memory bottleneck should be `P`, which is stored as `float32` and should thus consume around 903 MB of memory. However, the call to `impute` fails (see traceback below) as it requests `1.76TiB` of memory. That's because it tries to create an array of shape `Shape: f32[2039,17806,13298]`, which is not needed for this operation. 

Note that passing a batch size does not help much - let's say I'm passing `batch_size=500`, then this would still request an array of shape `2039 x 500 x 13298`, which still requires over 50GB of memory. Also, this this slows down solving the actual OT problem, which would not be necessary from a memory point of view. 

I talked to @michalk8 about this and it's probably a `vmap` that creates an array of the wrong shape. For now, we could solve this by evaluating the pull batch-wise over small sets of genes. This is inefficient, but would solve the issue for now. 

If the transport matrix fits into CPU memory, then the current best way to go about this is materializing the transport matrix before calling `impute`: 

```Python
for key, value in lmp.problems.items():
    value.solution.to(device="cpu")
    value.set_solution(np.array(value.solution.transport_matrix), overwrite=True)
```
That prevents the memory issue. 

```
2024-07-05 10:45:20.572529: W external[/tsl/tsl/framework/bfc_allocator.cc:485](http://localhost:53807/tsl/tsl/framework/bfc_allocator.cc#line=484)] Allocator (GPU_0_bfc) ran out of memory trying to allocate 1.76TiB (rounded to 1931211837440)requested by op 
2024-07-05 10:45:20.572824: W external[/tsl/tsl/framework/bfc_allocator.cc:497](http://localhost:53807/tsl/tsl/framework/bfc_allocator.cc#line=496)] *****_______________________________________________________________________________________________
[...]

PR #736, opened 2024-07-17
Fix/spatial mixins

Marius1311:
This implements a batched mode for `correlate` and `impute`, offering a temporary fix for https://github.com/theislab/moscot/issues/729.

### DIFF
--- a/src/moscot/problems/space/_mixins.py
+++ b/src/moscot/problems/space/_mixins.py
@@ -484,6 +484,7 @@ class SpatialMappingMixin(AnalysisMixin[K, B]):
         self: SpatialMappingMixinProtocol[K, B],
         var_names: Optional[Sequence[str]] = None,
         device: Optional[Device_t] = None,
+        batch_size: Optional[int] = None,
     ) -> AnnData:
         """Impute the expression of specific genes.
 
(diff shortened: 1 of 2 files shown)


src/moscot/problems/space/_mixins.py
```diff
@@ -414,10 +415,13 @@ def correlate(  # type: ignore[misc]
             - ``'pearson'`` - `Pearson correlation <https://en.wikipedia.org/wiki/Pearson_correlation_coefficient>`_.
             - ``'spearman'`` - `Spearman's rank correlation
                 <https://en.wikipedia.org/wiki/Spearman%27s_rank_correlation_coefficient>`_.
-        groupby
-            Optional `obs` field in `AnnData` to compute correlations over categorical groups.
         device
             Device where to transfer the solutions, see :meth:`~moscot.base.output.BaseSolverOutput.to`.
+        groupby
+            Optional `obs` field in `AnnData` to compute correlations over categorical groups.
```
MUCDK, line 421: I know this was there before but can we do ` `:attr:`~anndata.AnnData.obs ` `

MUCDK: :attr:`~anndata.AnnData.obs

MUCDK (changes requested):
Other than the minor doc thing everything looks good to me, thanks Marius!

and in the test maybe once test for this.

giovp (approved):

src/moscot/problems/space/_mixins.py
Marius1311: great, done!

Marius1311:
> Other than the minor doc thing everything looks good to me, thanks Marius!
> 
> and in the test maybe once test for this.

Thanks for your feedback @MUCDK, I added that minor change to the docs, extended tests to cover this case, spotted a corner case and fixed it. Everything should be fine now.

Issue #705, opened 2024-05-29
adapt low rank gamma default for linear problems


MUCDK:
for GW we keep default of 10.0, but for linear LR we need to set it to 500.

MUCDK:
also, when using LR, always set eps=0 per default, for both linear and quadratic problems.

PR #739, opened 2024-08-21
set defaults for lr sinkhorn

MUCDK:


MUCDK:
resolves #705

selmanozleyen (approved):
I'd think we should document the default gamma somewhere but it shouldn't matter as we are going to remove kwargs from the solve method in the future. So gamma can be explicitly documented there

Issue #716, opened 2024-06-14
Refactor: Maybe GW should inherit from FGW instead of vice versa


selmanozleyen:
I was wondering if it would make more sense to inherit `GWProblem` from `FGWProblem` since FGW kind of includes GW in itself. This is a problem since when GW inherites prepare it removes the linear arguments: (`xy_callback` etc.). And because of this we need to use `CompoundProblem.prepare` instead of `super().prepare` which is not ideal if you ask me.

 https://github.com/theislab/moscot/blob/d54060a83de43d779ac50e0125021f2904513e4d/src/moscot/problems/generic/_generic.py#L581-L592

MUCDK:
so you would then throw `NotImplementedError` for all functions which are specific to the fused term?

MUCDK:
Hi @selmanozleyen ,

Please give a brief update on this.

selmanozleyen:
I will close this as I noticed this is a bad idea when I was working on the solve args PR

PR #748, opened 2024-09-22
Refactor/handle solve args

selmanozleyen:
closes:
- https://github.com/theislab/moscot/issues/678
- https://github.com/theislab/moscot/issues/720

### DIFF
--- a/src/moscot/backends/ott/solver.py
+++ b/src/moscot/backends/ott/solver.py
@@ -456,6 +456,11 @@ class GWSolver(OTTJaxSolver):
             geom_kwargs["cost_matrix_rank"] = cost_matrix_rank
         geom_xx = self._create_geometry(x, t=time_scales_heat_kernel.x, is_linear_term=False, **geom_kwargs)
         geom_yy = self._create_geometry(y, t=time_scales_heat_kernel.y, is_linear_term=False, **geom_kwargs)
+        if alpha <= 0.0:
+            raise ValueError(f"Expected `alpha` to be in interval `(0, 1]`, found `{alpha}`.")
+        if (alpha == 1.0 and xy is not None) or (alpha != 1.0 and xy is None):
+            raise ValueError(f"Expected `xy` to be `None` if `alpha` is not 1.0, found xy={xy}, alpha={alpha}.")
+
         if alpha == 1.0 or xy is None:  # GW
             # arbitrary fused penalty; must be positive
             geom_xy, fused_penalty = None, 1.0
--- a/src/moscot/base/problems/problem.py
+++ b/src/moscot/base/problems/problem.py
@@ -430,24 +430,8 @@ class OTProblem(BaseProblem):
         solver_class = backends.get_solver(
             self.problem_kind, solver_name=solver_name, backend=backend, return_class=True
         )
-        init_kwargs, call_kwargs = solver_class._partition_kwargs(**kwargs)
-        # if linear problem, then alpha is 0.0 by default
-        # if quadratic problem, then alpha is 1.0 by default
-        alpha = call_kwargs.get("alpha", 0.0 if self.problem_kind == "linear" else 1.0)
-        if alpha < 0.0 or alpha > 1.0:
-            raise ValueError("Expected `alpha` to be in the range `[0, 1]`, found `{alpha}`.")
-        if self.problem_kind == "linear" and (alpha != 0.0 or not (self.x is None or self.y is None)):
-            raise ValueError("Unable to solve a linear problem with `alpha != 0` or `x` and `y` supplied.")
-        if self.problem_kind == "quadratic":
-            if self.x is None or self.y is None:
-                raise ValueError("Unable to solve a quadratic problem without `x` and `y` supplied.")
-            if alpha != 1.0 and self.xy is None:  # means FGW case
-                raise ValueError(
-                    "`alpha` must be 1.0 for quadratic problems without `xy` supplied. See `FGWProblem` class."
-                )
-            if alpha == 1.0 and self.xy is not None:
-                raise ValueError("Unable to solve a quadratic problem with `alpha = 1` and `xy` supplied.")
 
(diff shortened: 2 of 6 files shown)


src/moscot/backends/ott/solver.py
```diff
@@ -435,7 +435,7 @@ def _prepare(
         cost_matrix_rank: Optional[int] = None,
         time_scales_heat_kernel: Optional[TimeScalesHeatKernel] = None,
         # problem
-        alpha: float = 0.5,
+        alpha: Optional[float] = None,
```
MUCDK, line 438: can we just make a comment behind this that default is 0.5

MUCDK: I'm actually wondering whether we should have a default `alpha` whenever we don't want it to be 1.0

I.e. always set it explicitly in the classes which use GW, wdyt?

src/moscot/backends/ott/solver.py
```diff
@@ -456,6 +456,13 @@ def _prepare(
             geom_kwargs["cost_matrix_rank"] = cost_matrix_rank
         geom_xx = self._create_geometry(x, t=time_scales_heat_kernel.x, is_linear_term=False, **geom_kwargs)
         geom_yy = self._create_geometry(y, t=time_scales_heat_kernel.y, is_linear_term=False, **geom_kwargs)
+        if alpha is None:
+            alpha = 1.0 if xy is None else 0.5  # set defaults according to the data provided
+        if alpha <= 0.0:
```
MUCDK, line 461: also raise error if `alpha>1.0`

selmanozleyen: Actually https://github.com/theislab/moscot/blob/f303a14ac47b337b099b43800e567c89ca7b3dd3/src/moscot/backends/ott/_utils.py#L91 gives the error for invalid range. I think it's better if one source only throws error. And thinking of it I think even the check in 461 is redundant so I can remove it

src/moscot/backends/ott/solver.py
selmanozleyen: I set it like this so that it doesn't change the current behaviour. 
>I.e. always set it explicitly in the classes which use GW, wdyt?

You mean to make non-optional? I personally prefer non-optional parameters, especially if the class is an internal solver. I also think we should rename `GWSolver` to `FGWSolver` (because it technically can solve fgw and gw) and just set `alpha=1` when in a `GWProblem`.

MUCDK: I think that makes sense

src/moscot/backends/ott/solver.py
MUCDK: ok makes sense!

selmanozleyen:
Hi @MUCDK, except the notebooks I think it is ready to merge. But here are noteworthy side-effects that was caused by the ottjax update to 0.5.0

- `inner_iterations` parameter removed since it was fixed to 1 in new version of ottjax
- this is solved https://github.com/theislab/moscot/issues/716
- we no longer take `str` for initializers since ottjax dropped the support for this. We take initializer objects. I modified the tests accordingly but notebooks also need to comply to this

tests/_utils.py
```diff
@@ -101,3 +103,38 @@ def _base_problem_type(self) -> Type[B]:
...
+        return initializer
+    if initializer == "default":
+        return init_lib.DefaultInitializer(**kwargs)
+    if initializer == "gaussian":
+        return init_lib.GaussianInitializer(**kwargs)
+    if initializer == "sorting":
+        return init_lib.SortingInitializer(**kwargs)
+    if initializer == "subsample":
+        return init_lib.SubsampleInitializer(**kwargs)
+    raise NotImplementedError(f"Initializer `{initializer}` is not yet implemented.")
```
MUCDK, line 140: Can we keep strings for initializers , or do `Literal[...], init_lib.SinkhornInitializer` and port the code to src?

MUCDK (changes requested):
This looks great, thanks @selmanozleyen !

Wdyt about allowing strings for initialisers additionally? Might be too hard for people not familiar with these things otherwise.

selmanozleyen:
> This looks great, thanks @selmanozleyen !
> 
> Wdyt about allowing strings for initialisers additionally? Might be too hard for people not familiar with these things otherwise.

The issue there is differentiating the arguments required when creating the initializer objects vs the arguments required when calling initializers. For example you don't know if `rank` belongs to `RandomInitializer` or `solve(rank=rank)` in `initializer_kwargs` unless we have a hard coded list. We might not be able to use inspect module because some classes might be using `kwargs` in their constructors (because of this we won't know what the intializer takes). I will look if we can use inspect module but I think providing a class for creating initializers from strings is a better approach to clarify which argument goes where.

MUCDK:
> > This looks great, thanks @selmanozleyen !
> > Wdyt about allowing strings for initialisers additionally? Might be too hard for people not familiar with these things otherwise.
> 
> The issue there is differentiating the arguments required when creating the initializer objects vs the arguments required when calling initializers. For example you don't know if `rank` belongs to `RandomInitializer` or `solve(rank=rank)` in `initializer_kwargs` unless we have a hard coded list. We might not be able to use inspect module because some classes might be using `kwargs` in their constructors (because of this we won't know what the intializer takes). I will look if we can use inspect module but I think providing a class for creating initializers from strings is a better approach to clarify which argument goes where.

I see. Yes, then maybe let's create a class , but then call it from inside the `solve` method, because moscot users are not expected to know OOP. We are really supposed to do it in a simple manner.

selmanozleyen:
hi @MUCDK , I made the changes. If the user wants to create more advanced initializers (ie giving something to the initializer constructor as kwargs) they can do by creating it themselves. I think its reasonable to assume that someone who'd want to do that they can look ott initializers up.

MUCDK (approved):
That looks great, thanks so much @selmanozleyen !

MUCDK:
Please merge, and then do a new release v0.4.0 of moscot!

Issue #745, opened 2024-09-10
add geo link to moscot dataset pancreas


MUCDK:

MUCDK:
just add the link to the docstrings of `moscot.dataset.pancreas`

PR #750, opened 2024-10-05
Add link to pancreas dataset

LeonStadelmann:
#745


MUCDK:
Great, thanks Leon. Just realised that the pancreas dataset is not even in the docs yes, could you please add it, too?

MUCDK (approved):
Thanks!

PR #751, opened 2024-10-07
Fix `cell_transition` bug

selmanozleyen:
Hi @MUCDK ,

This used to give an error but now it doesn't
```python
import pandas as pd

adata_sc = datasets.drosophila(spatial=False)
adata_sp = datasets.drosophila(spatial=True)
adata_sc, adata_sp

if "test_col_1" in adata_sp.obs:
    del adata_sp.obs["test_col_1"]
if "test_col_2" in adata_sc.obs:
    del adata_sc.obs["test_col_2"]
if "test_col_1" in adata_sc.obs:
    del adata_sc.obs["test_col_1"]
if "test_col_2" in adata_sp.obs:
    del adata_sp.obs["test_col_2"]

adata_sc.obs["test_col_2"] = pd.Categorical(np.random.choice(["a", "b"], size=len(adata_sc)))
adata_sp.obs["test_col_1"] = pd.Categorical(np.random.choice(["a", "b"], size=len(adata_sp)))

mp = MappingProblem(adata_sc, adata_sp)
mp = mp.prepare(
    sc_attr={"attr": "obsm", "key": "X_pca"}, xy_callback="local-pca"
).solve(epsilon=10.0)


mp.cell_transition("src", "tgt", 
                   "test_col_1", "test_col_2")

```

I just separated the key used for storing the result and the key to push/pull.

I have some extra notes on function for future refactoring though:
- `other_keys` is always None from what I understand.
- Instead of a separate argument for cell aggregation I think user can just set `source_groups=None`.  (I am still not sure about this I am reading the code more atm)

### DIFF
--- a/src/moscot/base/problems/_mixins.py
+++ b/src/moscot/base/problems/_mixins.py
@@ -198,22 +198,24 @@ class AnalysisMixin(Generic[K, B]):
         )
         df_source = _get_df_cell_transition(
             self.adata,
-            [source_annotation_key] if aggregation_mode == "cell" else [source_annotation_key, target_annotation_key],
+            [source_annotation_key],
             key,
             source,
         )
         df_target = _get_df_cell_transition(
             self.adata if other_adata is None else other_adata,
-            [target_annotation_key] if aggregation_mode == "cell" else [source_annotation_key, target_annotation_key],
+            [target_annotation_key],
             key if other_adata is None else other_key,
             target,
         )
-
+        df_source = df_source.rename(columns={source_annotation_key: "res_annotation"})
+        df_target = df_target.rename(columns={target_annotation_key: "res_annotation"})
+        res_annotation_key = "res_annotation"
         source_annotations_verified, target_annotations_verified = _validate_annotations(
             df_source=df_source,
             df_target=df_target,
-            source_annotation_key=source_annotation_key,
-            target_annotation_key=target_annotation_key,
+            source_annotation_key=res_annotation_key,
+            target_annotation_key=res_annotation_key,
             source_annotations=source_annotations,
             target_annotations=target_annotations,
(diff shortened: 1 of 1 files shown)


src/moscot/base/problems/_mixins.py
```diff
@@ -198,22 +198,24 @@ def _cell_transition_online(
         )
         df_source = _get_df_cell_transition(
             self.adata,
-            [source_annotation_key] if aggregation_mode == "cell" else [source_annotation_key, target_annotation_key],
```
MUCDK, line 201: why can we drop the distinction between the two aggregation modes?

MUCDK:
I might miss something here, but why do you think that `other_key` is always `None`?
Whenever we have `other_adata`, we need it, don't we?
```
 _get_df_cell_transition(
            self.adata if other_adata is None else other_adata,
            [target_annotation_key] if aggregation_mode == "cell" else [source_annotation_key, target_annotation_key],
            key if other_adata is None else other_key,
            target,
        )
```

selmanozleyen:
Hi for example it's set to None here even though the other_data is given
https://github.com/theislab/moscot/blob/578e3eb85d5daab4f69356f15a73c66a8a4d6fea/src/moscot/problems/space/_mixins.py#L672-L685

I checked again and couldn't see.

selmanozleyen:
hi @MUCDK , I refactored the private methods with these objectives:
- less duplicate code
- less usage of pandas in main algorithms. ( I personally don't like using dataframes in algorithm implementations as it gets changed a lot and throws warnings or errors with it).  I only use pandas before the return line. 
- private methods take less arguments and are independent of `forwards` within themselves.

Would you like me to change anything else?

MUCDK (approved):
That looks good, thanks!

PR #755, opened 2024-10-10
update pancreas_multiome

MUCDK:


src/moscot/datasets.py
```diff
@@ -305,8 +306,8 @@ def pancreas_multiome(
     if rna_only:
         return _load_dataset_from_url(
             path,
-            file_type="h5ad",
-            backup_url="https://figshare.com/ndownloader/files/48785320",
+            type="h5ad",
```
selmanozleyen, line 309: @MUCDK  it's `file_type`, not `type`

MUCDK: ah great thanks, can you please adapt?

selmanozleyen (approved):
lgtm

Issue #765, opened 2024-11-20
Removing Protocol Classes


selmanozleyen:
Remove such classes and make reconfigure linter if you have to
```python
class AnalysisMixinProtocol(Protocol[K, B]):
    """Protocol class."""
    adata: AnnData
    _policy: SubsetPolicy[K]
    solutions: dict[tuple[K, K], BaseDiscreteSolverOutput]
    problems: dict[tuple[K, K], B]
```

PR #766, opened 2024-11-20
Refactor/remove protocol classes

selmanozleyen:
Solves [#765](https://github.com/theislab/moscot/issues/765)

There are some **cons** atm:

- I used Any  annotations in the abstract classes which is less specific than most of the protocols. See for example:
```python

class AbstractSolutionsProblems(ABC, metaclass=CombinedMeta):

    @property
    @abstractmethod
    def solutions(self) -> Any:
        """Solutions."""
        pass

    @property
    @abstractmethod
    def problems(self) -> Any:
        """Problems."""
        pass

    @property
    @abstractmethod
    def _policy(self) -> Any:
        """Subset policy."""
        pass
```
- For example one class uses src vs tgt terminology ( and other uses sp vs sc. They are essentially the same but I just rewrote them as is as two abstract classes. Otherwise it might also require more time to unify them.
These are the two classes I am talking about:
```python
class AbstractSrcTgt(ABC, metaclass=CombinedMeta):

    @property
    @abstractmethod
    def adata_src(self) -> AnnData:
        """Annotated data object."""
        pass

    @property
    @abstractmethod
    def adata_tgt(self) -> AnnData:
        """Annotated data object."""
        pass


class AbstractSpSc(ABC, metaclass=CombinedMeta):

    @property
    @abstractmethod
    def adata_sp(self) -> AnnData:
        """Annotated data object."""
        pass

    @property
    @abstractmethod
    def adata_sc(self) -> AnnData:
        """Annotated data object."""
        pass
```
- With the GENOT implementation `GENOTLinProblem` tries to inherit a class that requires `push/pull`
https://github.com/theislab/moscot/blob/f84f8be91e09a174edfbd1d4985e2dd575c0eafc/src/moscot/problems/generic/_generic.py#L779
**pros:**

- Less ignore comments for mypy
- Less code
- Better understanding of the expected attributes and methods

Things to do:
- Decide which files these new abstract classes should be in final version.
- Maybe fix the cons if important

### DIFF
--- a/src/moscot/base/problems/_mixins.py
+++ b/src/moscot/base/problems/_mixins.py
@@ -11,7 +11,6 @@ from typing import (
     Literal,
     Mapping,
     Optional,
-    Protocol,
     Sequence,
     Union,
 )
@@ -21,11 +20,9 @@ import pandas as pd
 from scipy.sparse.linalg import LinearOperator
 
 import scanpy as sc
-from anndata import AnnData
 
 from moscot import _constants
 from moscot._types import ArrayLike, Numeric_t, Str_Dict_t
-from moscot.base.output import BaseDiscreteSolverOutput
 from moscot.base.problems._utils import (
     _check_argument_compatibility_cell_transition,
     _correlation_test,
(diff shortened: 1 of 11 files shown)


MUCDK (approved):
This is amazing, thanks @selmanozleyen

PR #778, opened 2025-01-07
Neural Updates: Move GENOTLinProblem, Update SolverOutput Class Hierarchy

selmanozleyen:
Important things to note:
- removed this class from the user.rst file because I think it should be tested more to be exposed in the docs
- In `project_to_transport_matrix` I removed the `forward` argument since for that class forward is always true.
- In general I added a method `_apply_forward` to the base output class so it includes both the neural and non-neural case. This might seem like more code but checking every time for forward=False and throwing an error is probably more problematic and has more code. For the old moscot classes nothing changes. `_apply_forward` is also simply supported with `def _apply_forward(x): return apply(x, forward=True)`


MUCDK (approved):
Thanks!

Issue #811, opened 2025-03-19
Fix AlignmentProblem


MUCDK:
Following updates of some dependencies, the `AlignmentProblem` does cause some issues, see https://github.com/theislab/moscot/issues/806, https://github.com/theislab/moscot/issues/801, https://github.com/theislab/moscot/issues/798 .

Let's fix this, and clarify why our tests did not fail.

Issue #801, opened 2025-02-19
spatial AlignmentProblem for spatial omics slides with different number of cells shuffles cells randomly


hspitzer:
Hi,

I've been experimenting with this really nice package for the last few days, for aligning spatial transcriptomics and spatial proteomics slides. First within modality, but I'm also keen to experiment with multimodal alignment. I have come across a potential bug, that might be related to #798 where the cells seem to be completely shuffled after alignment. 

This only occurs when the shape of the different slides that I am aligning are not the same. In addition, I get an Error about incompatible shapes when trying to align slides with the "star" policy. I believe these two things might be related, so I am putting both of them in the same issue, but happy to split them up if more convenient. 

I would be thankful for any pointers or help on this!

Here a reproducible example (adapted from the [spatial alignment example](https://moscot.readthedocs.io/en/latest/notebooks/tutorials/300_spatial_alignment.html):

```python
from moscot import datasets
from moscot.problems.space import AlignmentProblem
import scanpy as sc
import squidpy as sq

adata = datasets.sim_align()
# add x coordinate to obs to visualise it later on
adata.obs['x'] = adata.obsm['spatial'][:, 0]
# subsample adata to get slightly different number of cells for every batch
# if we don't do this, the errors don't ocurr and everything behaves normally
sc.pp.subsample(adata, fraction=0.99)
display(adata.obs.groupby('batch').count())
sq.pl.spatial_scatter(adata, shape=None, library_id="batch", color=["batch", 'x'])
```

This results in the following image:

![Image](https://github.com/user-attachments/assets/0b4f9a84-b975-46f6-b21e-8bde89561338)

```python
ap = AlignmentProblem(adata=adata)
ap = ap.prepare(batch_key="batch", policy="sequential")
ap = ap.solve()
ap.align(key_added="spatial_warp",  reference='0')
sq.pl.spatial_scatter(
    adata, shape=None, spatial_key="spatial_warp", library_key='batch', color=["batch", 'x']
)
```
Now, after the alignment all x coordinates are completely shuffled, even though the points look like they are aligned nicely

![Image](https://github.com/user-attachments/assets/cb0cfd25-0a21-484d-9c40-d77a7ebe0043)

Secondly, here the error that ocurrs when trying the "star" policy
```python
ap = AlignmentProblem(adata=adata)
ap = ap.prepare(batch_key="batch", policy="star", reference='0')
ap = ap.solve()
ap.align(key_added="spatial_warp",  reference='0')
```
<details>
[...]

Issue #798, opened 2025-02-10
Chaotic Spatial


DruncBread:
Hi,

I tried moscot on two DLPFC slices 151673 and 151507 ( 151673 as reference).
But after running the codes below for alignment, I found that the resulted spatial is messed up.
Would you mind helping me solve this problem?
Thank you!

```
adata = ad.concat([target,source])

ap = AlignmentProblem(adata=adata)
ap = ap.prepare(batch_key="batch", policy="sequential")
ap = ap.solve()

mode='affine'
ap.align(reference='151673' , mode=mode, spatial_key="spatial", key_added=f"spatial_{mode}")
sq.pl.spatial_scatter(
    adata, shape=None, spatial_key=f"spatial_{mode}", library_id="batch", color="batch"
)

```

I visualized the slices and found that the domain is messed up. 

![Image](https://github.com/user-attachments/assets/73938e03-dc81-465c-b275-b810a807cc3b)

![Image](https://github.com/user-attachments/assets/044274b3-a1f8-4811-bcf4-625f76d8b7fc)

DruncBread:
This is my datasets and the code to load the data

DruncBread:
Hi, 

Sorry about the unclear description! 

So I am attempting to align two DLPFC slices (151507 to 151673) using affine transformation. ( 151673 as reference, 151507 aligned to 151673 ).

![Image](https://github.com/user-attachments/assets/17020819-4b6d-4c4c-a734-94d4a02351cd)

![Image](https://github.com/user-attachments/assets/23678722-5587-42e7-a300-06aa94b2d618)

However, after running the code above, the result was not as expected

As you can see in the code, I used affine transformation. So the aligned slices should have been simply rotated, but instead, the spots' domain (obs['ground_truth']) is incorrect, and it seems that the spots have been mixed up in the transformation.

![Image](https://github.com/user-attachments/assets/499a087d-b9a3-4d93-8537-edde62cf161e)

![Image](https://github.com/user-attachments/assets/57dc5115-9469-4271-96b5-9fe0bee66b24)

![Image](https://github.com/user-attachments/assets/658a1d6a-9cac-4427-9032-849e7c0d27b9)


I failed to upload the DLPFC dataset files on github. Would you mind leave your email so I could send you the data?

Thank you!

Best regards.

PR #814, opened 2025-03-21
Align method assignment fix

selmanozleyen:
Turns out the bug was because of the way the results are assigned to adata.

Fixes https://github.com/theislab/moscot/issues/811, https://github.com/theislab/moscot/issues/801, https://github.com/theislab/moscot/issues/798 and adds tests to it. The test fails on the unfixed version. You can also checked the corrected visualization from my comment https://github.com/theislab/moscot/issues/801#issuecomment-2743353442

The problem was the assignment.
```python
aligned_basis = np.vstack([aligned_maps[k] for k in self._policy._cat])
# the order of adata and the order of self._policy didn’t match
self.adata.obsm[key_added] = aligned_basis
```
So in this PR I made sure to assing the adata by the batches correctly


tests/problems/space/test_alignment_problem.py
```diff
@@ -235,3 +235,23 @@ def test_pass_arguments(self, adata_space_rotate: AnnData, args_to_check: Mappin
         for arg, val in pointcloud_args.items():
             assert hasattr(geom, val)
             assert getattr(geom, val) == args_to_check[arg]
+
+    def test_alignment_order_preservation(self, adata_space_rotate: AnnData):
```
MUCDK, line 239: Can we have this test for all policies which we allow for the `AlignmentProblem`?

tests/problems/space/test_alignment_problem.py
```diff
@@ -235,3 +235,23 @@ def test_pass_arguments(self, adata_space_rotate: AnnData, args_to_check: Mappin
...
+        ap.align(key_added="spatial_warped", mode="warp", reference="0")
+
+        threshold = 0.7
+        batch = "0"
+        mask = ap.adata.obs["batch"] == batch
+        aligned_x = ap.adata.obsm["spatial_warped"][mask, 0]
+        rank_aligned_x = np.argsort(aligned_x)
+        original_x = ap.adata.obsm["spatial"][mask, 0]
+        rank_original_x = np.argsort(original_x)
+        corr = np.corrcoef(rank_original_x, rank_aligned_x)[0, 1]
```
MUCDK, line 253: what is the actual value for this example? then we can make it potentially tighter, thus closer to a regression test.

MUCDK (changes requested):
Thanks a lot @selmanozleyen !

let's make sure to test this a bit better to not encounter the issue again

tests/problems/space/test_alignment_problem.py
selmanozleyen: It is 0.9999999999999999

MUCDK: oh wow, isn't that suspiciously high?

tests/problems/space/test_alignment_problem.py
selmanozleyen: ok added star with reference 0 and 1, and set the threshold to 0.96.

tests/problems/space/test_alignment_problem.py
selmanozleyen: But I only check on the reference node. In fact I realized it should in fact stay the same maybe I should change it to equality then. And I checked they are equal at the moment. This is how we want the end result to look right? https://github.com/theislab/moscot/issues/801#issuecomment-2743353442

selmanozleyen: Ok I updated the tests now it checks if the order is preserved or not. I also set `max_iterations=2` to speed it up since I realized this bug wasn't related to the transport matrix

MUCDK: and yeah, I do agree that the reference batch remains the same, let's test for equality of the reference slide, and let's test for a high correlation for aligned slices.

selmanozleyen: thats correct let me do that

MUCDK: can we fix this please?

selmanozleyen:
I noticed ranked correlation is a bad idea for checking other batches. For more general tests I computed the correlation of flattened distance matrices. They seem to make sense so far. One thing to note is that when the solver doesn't converge the correlation results are very high so I also assert for convergence.

Here I can show the results of the test data
```python
print(f"TESTING ALIGNMENT ORDER PRESERVATION FOR {policy} POLICY AND REFERENCE {reference}")
for batch in adata_space_rotate.obs["batch"].unique():
    mask = adata_space_rotate.obs["batch"] == batch
    aligned_coords = adata_space_rotate.obsm["spatial_warped"][mask]
    original_coords = adata_space_rotate.obsm["spatial"][mask]
    original_dist = squareform(pdist(original_coords)).ravel()
    aligned_dist = squareform(pdist(aligned_coords)).ravel()

    distance_correlation = np.corrcoef(original_dist, aligned_dist)[0, 1]
    mse = np.mean((original_dist - aligned_dist) ** 2)

    print(f"> Batch {batch}, distance correlation: {distance_correlation}, MSE: {mse}")
            print(f"> Batch {batch}, distance correlation: {distance_correlation}, MSE: {mse}")
```
Here is the corresponding output
```
TESTING ALIGNMENT ORDER PRESERVATION FOR star POLICY AND REFERENCE 0
> Batch 2, distance correlation: 0.843068707925429, MSE: 36.938263347797914
> Batch 1, distance correlation: 0.8424264059465704, MSE: 36.90107597359456
> Batch 0, distance correlation: 1.0, MSE: 0.0

TESTING ALIGNMENT ORDER PRESERVATION FOR star POLICY AND REFERENCE 1
> Batch 2, distance correlation: 0.8432435178608674, MSE: 36.9919495766207
> Batch 1, distance correlation: 1.0, MSE: 0.0
> Batch 0, distance correlation: 0.843847004686296, MSE: 37.17927760765656

TESTING ALIGNMENT ORDER PRESERVATION FOR sequential POLICY AND REFERENCE 0
> Batch 2, distance correlation: 0.6225322935621641, MSE: 40.51853053060476
> Batch 1, distance correlation: 0.8424264059464235, MSE: 36.901075973594516
> Batch 0, distance correlation: 1.0, MSE: 0.0
```

So here is the latest version

```python
threshold = 0.6
sc.pp.subsample(adata_space_rotate, fraction=0.99)
ap: AlignmentProblem = AlignmentProblem(adata=adata_space_rotate)
ap = ap.prepare(batch_key="batch", joint_attr={"attr": "X"}, policy=policy, reference=reference)

ap = ap.solve(alpha=0.5, epsilon=1, rank=-1)
assert np.all([sol.converged for sol in ap.solutions.values()])

ap.align(key_added="spatial_warped", mode="warp", reference=reference)
for batch in adata_space_rotate.obs["batch"].unique():
    mask = adata_space_rotate.obs["batch"] == batch
    aligned_coords = adata_space_rotate.obsm["spatial_warped"][mask]
    original_coords = adata_space_rotate.obsm["spatial"][mask]
    original_dist = squareform(pdist(original_coords)).ravel()
    aligned_dist = squareform(pdist(aligned_coords)).ravel()

    distance_correlation = np.corrcoef(original_dist, aligned_dist)[0, 1]
    assert distance_correlation > threshold, f"Batch {batch}, distance correlation: {distance_correlation}"
    if reference == batch:
        assert distance_correlation == 1.0, (
            f"The reference batch {batch} should not be warped, correlation: {distance_correlation}"
        )
    else:
        assert distance_correlation < 1.0, f"Batch {batch}, distance correlation: {distance_correlation}"

```

MUCDK:
ok great, that looks good, thanks so much, this looks great!

MUCDK (approved):
Amazing thanks so much @selmanozleyen !

MUCDK:
@selmanozleyen feel free to merge and make a new patch release of moscot.

Issue #832, opened 2025-06-27
Fixes of moscot code


MUCDK:
- update to JAX 0.6.1 https://github.com/theislab/moscot/issues/830
- following https://github.com/theislab/moscot/issues/813, for the moment include these dependencies as core dependency. only later remove

PR #833, opened 2025-06-29
Fix jax version and neural

### LINKED ISSUE #830: current moscot release is not compatible with Jax 0.6.1
In the following PR https://github.com/jax-ml/jax/pull/28181 the `xla_extension `was renamed to `_jax` causing issues when trying to use the most recent moscot release (0.4.2): `ModuleNotFoundError: No module named 'jaxlib.xla_extension'`

The issue can be fixed by installing jax<=0.6.0 or installing moscot directly from main as the import line causing the error in src/moscot/backends/ott/output.py  was removed in commit 8b1a9c5.

Since the problem seems to be addressed mainly posting for any other users stumbling across this.

LeonStadelmann:
Addressing #832.
I added a requirement for the jax version to be >=0.6.1. If I understand correctly, this does not fix #830 though since the error appears when installing moscot with pip. So I think the error will still occur in the pip-downloaded version until the next release.

I moved all the requirements for moscot neural into the core dependencies to make their import not optional anymore.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,11 +54,16 @@ dependencies = [
     "scanpy>=1.9.3",
     "wrapt>=1.13.2",
     "docrep>=0.3.2",
+    "jax>=0.6.1",
     "ott-jax>=0.5.0",
     "cloudpickle>=2.2.0",
     "rich>=13.5",
     "docstring_inheritance>=2.0.0",
-    "mudata>=0.2.2"
+    "mudata>=0.2.2",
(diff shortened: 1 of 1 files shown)


MUCDK:
Thanks @LeonStadelmann 

Do you mean that when installing with `pip` you still get the error? If so, this is due to us having to publish a new release on pypi, which we will do after merging this PR

MUCDK (approved):

